PR #78, opened 2020-06-14
Feature/1 multi level params

simonlousky:
log_hyperparams now saves a nested param yaml file, since multi-level params are now supported on [DAGsHub](https://dagshub.com/)

### DIFF
--- a/dagshub/logger.py
+++ b/dagshub/logger.py
@@ -69,7 +69,7 @@ class DAGsHubLogger:
     def log_hyperparams(self, params: Dict[str, Any] = None, **kwargs):
         if self.should_log_hparams:
           self.hparams.update(self.normalize_dictionary_values(params) or {})
-          self.hparams.update(kwargs)
+          self.hparams.update(self.normalize_dictionary_values(kwargs or {}))
           if self.eager_logging:
               self.save_hparams()
 
--- a/dagshub/logger.py
+++ b/dagshub/logger.py
@@ -3,7 +3,6 @@ from contextlib import contextmanager
 from datetime import datetime
 from typing import TextIO, ContextManager, Any, Dict
 
-import copy
 import yaml
 import csv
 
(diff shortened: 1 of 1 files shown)


dagshub/logger.py
```diff
@@ -68,10 +68,10 @@ def epoch_milisec():
 
     def log_hyperparams(self, params: Dict[str, Any] = None, **kwargs):
         if self.should_log_hparams:
-            self.hparams.update(params or {})
-            self.hparams.update(kwargs)
-            if self.eager_logging:
-                self.save_hparams()
+          self.hparams.update(self.normalize_dictionary_values(params) or {})
+          self.hparams.update(kwargs)
```
guysmoilov, line 72: Why not normalize kwargs?
```suggestion
          self.hparams.update(self.normalize_dictionary_values(kwargs or {}))
```

dagshub/logger.py
```diff
@@ -68,10 +68,10 @@ def epoch_milisec():
 
     def log_hyperparams(self, params: Dict[str, Any] = None, **kwargs):
         if self.should_log_hparams:
-            self.hparams.update(params or {})
-            self.hparams.update(kwargs)
-            if self.eager_logging:
-                self.save_hparams()
+          self.hparams.update(self.normalize_dictionary_values(params) or {})
```
guysmoilov, line 71: I think it makes more sense to check if params is None and this will automatically make `normalize_dictionary_values` never return None
```suggestion
          self.hparams.update(self.normalize_dictionary_values(params or {}))
```

dagshub/logger.py
```diff
@@ -93,6 +93,23 @@ def save_hparams(self):
...
+        for key, value in dictionary.items():
+          if isinstance(value, dict):
+            new_dict[key] = normalize_dict_deep(value)
+            continue
+          new_dict[key] = value if value is None or type(value) in [int, float, bool] else str(value)
+
+        return new_dict
+
+      return normalize_dict_deep(dictionary)
+
```
guysmoilov, line 112: This can be turned into just one function

simonlousky: I thought about that and I prefer not having a method in the class that is itself recursive.
It's a better practice to keep the recursive logic in an inner function

PR #91, opened 2020-09-06
First pull request for fastai autologger

deanp70:


requirements.txt
```diff
@@ -65,3 +65,7 @@ urllib3==1.25.10
 wcwidth==0.2.5
 Werkzeug==1.0.1
 zipp==3.1.0
+
+fastai~=2.0.8
+fastai2~=0.0.30
+setuptools~=47.1.0
```
guysmoilov, line 71: Is this required by fastai? Why add it now?

requirements.txt
```diff
@@ -65,3 +65,7 @@ urllib3==1.25.10
 wcwidth==0.2.5
 Werkzeug==1.0.1
 zipp==3.1.0
+
+fastai~=2.0.8
+fastai2~=0.0.30
```
guysmoilov, line 70: Two versions of fastai?

dagshub/fastai/README.md
```diff
@@ -0,0 +1,25 @@
+# [Fastai v2](https://github.com/fastai/fastai) + [DAGsHub](https://dagshub.com) integration
+
+This package allows you to output logs from `fastai` runs to a simple, open format used by DAGsHub.
```
guysmoilov, line 3:
```suggestion
This package allows you to output logs from `fastai` runs to a [simple, open format](https://dagshub.com/docs/reference/open_data_science_formats/), that you can use analyze yourself or display using DAGsHub.
```

dagshub/fastai/logger.py
```diff
@@ -0,0 +1,84 @@
...
+        self._dags_status_hyperparam_name = 'success'
+        self._dags_step_num = -1
+        self._dags_epoch = 0
+
+    def before_fit(self):
+        # Make sure this is a training run
+        self.run = not hasattr(self.learn, 'lr_finder') and \
+                   not hasattr(self, "gather_preds") and rank_distrib() == 0
+        if not self.run:
+            return
```
guysmoilov, line 51: What does it mean if it's no a training run? Inference?
Also, `self.run` is a misleading name.
`self._is_training_run`?

guysmoilov: And, is this check an undocumented hack? Can fastai change these field names without notice?

dagshub/fastai/logger.py
```diff
@@ -0,0 +1,84 @@
...
+
+        # Log Hyper Parameters
+        params = self.learn.gather_args()
+        self.logger.log_hyperparams(params)
+
+    def after_batch(self):
+        if self.learn.training:
+            self._dags_step_num += 1
```
guysmoilov, line 59: Maybe we should re-zero `_dags_step_num` in `before_fit` or something? Is there risk of the logger being reused across runs (when used appropriately)?

dagshub/fastai/logger.py
```diff
@@ -0,0 +1,84 @@
...
+                   not hasattr(self, "gather_preds") and rank_distrib() == 0
+        if not self.run:
+            return
+
+        # Log Hyper Parameters
+        params = self.learn.gather_args()
+        self.logger.log_hyperparams(params)
+
+    def after_batch(self):
+        if self.learn.training:
```
guysmoilov, line 58: I'm confused. If this flag exists, what's with the `self.run` check?

dagshub/fastai/logger.py
```diff
@@ -0,0 +1,84 @@
...
+            self._dags_step_num += 1
+            hypers = {f'{k}_{i}': v for i, h in enumerate(self.opt.hypers) for k, v in h.items()}
```
guysmoilov, line 60: What's hypers?

dagshub/fastai/logger.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def after_epoch(self):
+        self._dags_epoch += 1
+        metrics = {n: s for n, s in zip(self.recorder.metric_names, self.recorder.log) if n not
+                   in ['train_loss', 'epoch', 'time']}
+        metrics['epoch'] = self._dags_epoch
+        self.logger.log_metrics(metrics, step_num=self._dags_step_num)
+
+    def after_fit(self):
+        self.run = True
```
guysmoilov, line 76: Why? Doesn't it mean that the run is actually over?

dagshub/fastai/logger.py
deanp70: I think this is originally borrowed from the tensorboard tracking notebook, and yes, it seems like a hack. I don't know what else we can do about it though. 

Re: changing the name, I'll do it.

requirements.txt
deanp70: I checked and the fastai2 is not needed. Wierd behavior though

requirements.txt
deanp70: Removed

dagshub/fastai/logger.py
deanp70: So upon further research `self.run` is a reserved attribute, and it complains if I rename it to something else. From the code, `self.run` should also be reset to True after each fit, which is why it exists in the code at the end.

dagshub/fastai/logger.py
deanp70: This checks that the learner is in the process of training, i.e. not that this is a training run, but that we are in the middle of the training process.

dagshub/fastai/logger.py
deanp70: Not sure what to tell you. Sometimes Jeremy recommends fitting for 1 cycle and then fitting for another. In such a case you'd probably not want to zero the steps. This is also the practice in other loggers.

dagshub/fastai/logger.py
deanp70: Again, this is a bad name that is used througout all of fastai. Apparantly these are the optimizer hyperparams, which might change throughout the training loop. In the example I ran on my system (Savta), it seems the lr is adjusted for example. It seems strange that it must be logged as a metric, but since it changes over time it might make sense.

dagshub/fastai/logger.py
deanp70: Required by fastai, as I wrote above.

dagshub/fastai/logger.py
guysmoilov: For the record, its meaning is "should this callback even be called"

PR #111, opened 2021-01-14
Update logger.py

deanp70:


dagshub/fastai/logger.py
```diff
@@ -17,6 +21,9 @@ class DAGsHubLogger(Callback):
     learn.fit(..., cbs=DAGsHubLogger())
     ```
     """
+    
+    "Saves model topology, losses & metrics"
+    remove_on_fetch, order = True, Recorder.order+1
```
deanp70, line 26: This is FastAI black magic. Not sure what it is and why it's necessary, but you must have this otherwise logging doesn't work

dagshub/fastai/logger.py
```diff
@@ -1,4 +1,8 @@
+from fastai.learner import Learner, Recorder
+from fastcore.basics import *
+from fastai.callback.progress import *
```
deanp70, line 3:
```suggestion
```

PR #119, opened 2021-07-26
Logger PyTorch lightning import fix 

gagan3012:
Hello,
I was working with Pytorch lightning and the logger is out of date
This PR updates the logger to work with Pytorch-lightning

PR #120, opened 2021-07-28
Update for recent version of PyTorch Lightning

deanp70:
Continues the work started in #119.


deanp70:
This isn't working yet. Still throws some errors

PR #121, opened 2021-07-29
Fix examples, readme and the logger

mohithg:


dagshub/pytorch_lightning/logger.py
```diff
@@ -35,10 +38,28 @@ def __init__(self,
         self.logger = LoggerImpl(metrics_path=metrics_path, should_log_metrics=should_log_metrics,
                                  hparams_path=hparams_path, should_log_hparams=should_log_hparams,
                                  should_make_dirs=should_make_dirs, eager_logging=False)
+        self._name = name or ""
+        self._version = version
+        self._experiment = None
```
deanp70, line 43: I guess this isn't necessary if we are going to always return the `self.logger` in the `experiment` function.

dagshub/pytorch_lightning/logger.py
```diff
@@ -35,10 +38,28 @@ def __init__(self,
...
+    @rank_zero_experiment
+    def experiment(self) -> LoggerImpl:
+        r"""
+        Actual ExperimentWriter object. To use ExperimentWriter features in your
+        :class:`~pytorch_lightning.core.lightning.LightningModule` do the following.
+        Example::
+            self.logger.experiment.some_experiment_writer_function()
+        """
+        if self._experiment:
+            return self._experiment
```
deanp70, line 56: Same for these 2 lines, we can remove them I think

deanp70 (changes requested):

deanp70 (approved):

deanp70 (changes requested):
Can you please add to the general readme in the Integrations section for pytorch lightning:
`supports version 1.4.0 or higher`

deanp70 (approved):

PR #183, opened 2022-09-28
Added OAuth token handling

kbolashev:
Tokens are cached on disk in `{user_cache}/dagshub/tokens` file, keyed by host.

Users (and the client library for that matter) should be using `dagshub.auth.get_token()` and it will do everything for them.
Tokens are cached at runtime too, so `get_token()` can be called multiple times with only overhead being the expiration check.
Host, client_id and cache location are overridable with env vars


kbolashev:
[OAuth authentication in the DagsHub client](https://trello.com/c/UmHS3Vws/1377-oauth-authentication-in-the-dagshub-client)

setup.py
```diff
@@ -23,6 +23,7 @@
     entry_points={
         "console_scripts": [
             "dagshub-mount = dagshub.streaming.mount:main",
+            "dagshub-auth = dagshub.auth:console_entrypoint",
```
guysmoilov, line 26: The result command line would be `dagshub-auth`? Any easy option to make it `dagshub login`?

dagshub/auth/oauth.py
```diff
@@ -0,0 +1,126 @@
...
+        _authenticator = OAuthAuthenticator(**kwargs)
+    return _authenticator.get_oauth_token()
+
+
+def console_entrypoint():
+    get_oauth_token()
+    print(
+        f"Your DagsHub OAuth token is now stored at {_authenticator.cache_location}"
+        f" and will be used next time you use the client"
+    )
```
guysmoilov, line 120: I'm not sure whether this is TMI or not

dagshub/auth/oauth.py
```diff
@@ -0,0 +1,126 @@
...
+            self._store_cache_file()
+        return token["access_token"]
+
+    def oauth_flow(self) -> Dict:
+        state = uuid.uuid4()
+        dagshub_url = urllib.parse.urljoin(self.host, "login/oauth")
+        print(
+            f"Go to {dagshub_url}/authorize?state={state}&client_id={self.client_id} and paste the code back in here."
+        )
+        code = getpass.getpass("Code:")
```
guysmoilov, line 50: Since this code SHOULD get executed in CI\CD scenarios, is there a standard way for you to check whether you're in an interactive session, and if not then just crash instead of getting the machine stuck on an input and wasting user cloud credits?

dagshub/auth/oauth.py
```diff
@@ -0,0 +1,126 @@
...
+class OAuthAuthenticator:
+    def __init__(self, **kwargs):
+        self.host: str = kwargs.get("host", os.environ.get(HOST_KEY, DEFAULT_HOST)).strip("/")
+        self.client_id: str = kwargs.get(
+            "client_id", os.environ.get(CLIENT_ID_KEY, DEFAULT_CLIENT_ID)
+        )
+        self.cache_location: str = kwargs.get(
+            "cache_location", os.environ.get(CACHE_LOCATION_KEY, DEFAULT_CACHE_LOCATION)
+        )
+        self._token_cache: Optional[Dict[str, Dict]] = None
```
guysmoilov, line 32: Why this `kwargs` based pattern? Won't it be more convenient to have named args with default values?
I know the default depend on env vars, but those rarely change during runtime, so should be fine to define them at class creation

guysmoilov (changes requested):
Partial review

dagshub/auth/oauth.py
kbolashev: This is purely my laziness because I like chaining .get()s this way, and you can't do that with named args.
It's worse for readability/autocomplete, so I should probably go with named args, you're right

dagshub/auth/oauth.py
kbolashev: Will research more

dagshub/auth/oauth.py
guysmoilov: Why not actually? Can't you do this:
```python
def foo(bar = os.environ.get('XXX', 'YYY'))
```

dagshub/auth/tokens.py
```diff
@@ -0,0 +1,126 @@
...
+        if host not in self._token_cache:
+            self._token_cache[host] = []
+        self._token_cache[host].append(token)
+        self._store_cache_file()
+
+    def get_token(self, host: str = None):
+        if host is None:
+            host = config.host
+        tokens = self._token_cache.get(host, [])
+        app_tokens = list(filter(lambda t: t["token_type"] == APP_TOKEN_TYPE, tokens))
```
simonlsk, line 42: I prefer list comprehension for a simple `==` comparison, but that's a matter of taste, you can leave it like that. 
Use `t.get("token_type")` though to prevent potential `KeyError` exceptions.

dagshub/auth/tokens.py
```diff
@@ -0,0 +1,126 @@
...
+        if len(app_tokens) > 0:
+            token = app_tokens[0]
```
simonlsk, line 44: If I manually invalidate a token on DagsHub, is the latest going to be promoted to the top of the list? If not maybe use `-1`

dagshub/auth/tokens.py
```diff
@@ -0,0 +1,126 @@
...
+        else:
+            non_expired_tokens = list(filter(lambda t: not self._is_expired(t), tokens))
+            if len(non_expired_tokens) > 0:
+                token = non_expired_tokens[0]
```
simonlsk, line 48: Same comment here

dagshub/auth/tokens.py
```diff
@@ -0,0 +1,126 @@
...
+            else:
+                logger.warning(f"No valid tokens left for host '{host}'. Reauthenticating with OAuth")
```
simonlsk, line 50:
```suggestion
                logger.warning(f"No valid tokens found for host '{host}'. Authenticating with OAuth")
```
What if this is the first time?

dagshub/auth/tokens.py
```diff
@@ -0,0 +1,126 @@
...
+        expiry = token["expiry"][:-4] + "Z"
+        expiry_dt = datetime.datetime.strptime(expiry, "%Y-%m-%dT%H:%M:%S.%fZ")
+        is_expired = expiry_dt < datetime.datetime.utcnow()
+        return is_expired
+
+    def _load_cache_file(self) -> Dict[str, List[Dict]]:
+        logger.debug(f"Loading OAuth token cache from {self.cache_location}")
+        if not os.path.exists(self.cache_location):
+            logger.debug("OAuth token cache file doesn't exist")
+            return {"version": config.CONFIG_SCHEMA_VERSION}
```
simonlsk, line 73: I somehow misinterpreted this as an error response (probably because of the log message just above) at the beginning. Add a const `EMPTY_CACHE_DICT` or something like that.

dagshub/auth/tokens.py
```diff
@@ -0,0 +1,126 @@
...
+            raise
+
+    def _store_cache_file(self):
+        logger.debug(f"Dumping OAuth token cache to {self.cache_location}")
+        try:
+            dirpath = os.path.dirname(self.cache_location)
+            if not os.path.exists(dirpath):
+                os.makedirs(dirpath)
+            with open(self.cache_location, "w") as f:
+                yaml.dump(self.__token_cache, f, yaml.Dumper)
```
simonlsk, line 91: Not critical now I guess, but this loading and saving flow is not safe for parallel work on the same cache file. The latest to save will override whatever is in there.

dagshub/common/config.py
```diff
@@ -0,0 +1,14 @@
+import appdirs
+import os
+
+HOST_KEY = "DAGSHUB_CLIENT_HOST"
+DEFAULT_HOST = "https://dagshub.com"
+CLIENT_ID_KEY = "DAGSHUB_CLIENT_ID"
+DEFAULT_CLIENT_ID = "32b60ba385aa7cecf24046d8195a71c07dd345d9657977863b52e7748e0f0f28"
```
simonlsk, line 7: Where does that come from? Do we have a UI for that in DagsHub? Just curious.

dagshub/common/config.py
```diff
@@ -0,0 +1,14 @@
...
+CACHE_LOCATION_KEY = "DAGSHUB_CLIENT_CONFIG"
+DEFAULT_CACHE_LOCATION = os.path.join(appdirs.user_cache_dir("dagshub"), "tokens")
+CONFIG_SCHEMA_VERSION = "1"
```
simonlsk, line 10: Clarify `config` vs `tokens` terminology to me. If we decide to actually save configs for the client, are we going to use the `user_cache_dir` as well? 
If we introduce a new `"dagshub-config"` file, does the schema change? Or does it refer to the `"tokens"` schema?

dagshub/common/cli.py
```diff
@@ -0,0 +1,73 @@
...
+
+@auth.command(name="add")
+@click.argument("token")
+@click.pass_context
+def auth_add(ctx, token):
+    """
+    Add a long-lived auth token
+    """
+    dagshub.auth.add_app_token(token, ctx.obj["host"])
+    print("Token added")
```
simonlsk, line 59: I like the `gcloud` / `docker` / `heroku` style:
`dagshub login` will actually trigger oauth. 
Beside that, maybe regular long-lived tokens shouldn't even be cached, just managed by the user with whatever secret system they want. They would need to pass it as an argument to whatever function they use.
For CLI mount that would need to be
```bash 
dagshub mount --username <user> --password <pass>
```
WDYT?

simonlsk (changes requested):

dagshub/auth/oauth.py
```diff
@@ -0,0 +1,56 @@
...
+    if code_input_timeout is None:
+        code_input_timeout = CODE_INPUT_TIMEOUT
+    state = uuid.uuid4()
+    host = host.strip("/")
+    dagshub_url = urllib.parse.urljoin(host, "login/oauth")
+    link_prompt = f"Go to {dagshub_url}/authorize?state={state}&client_id={client_id} and paste the code back in here."
+    code_prompt = "Code:"
+    if code_input_timeout <= 0:
+        print(link_prompt)
+        code = input(code_prompt)
```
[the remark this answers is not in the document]
kbolashev: The redeem code is short-lived and supposed to be (but not sure if it is) a once-use only.
I will put a getpass in here though, you're right.
I think I used an input here because I didn't like how it shows you warnings in IntelliJ when you use it, but that's less relevant than security

dagshub/auth/oauth.py
```diff
@@ -0,0 +1,56 @@
...
+    if client_id is None:
+        client_id = config.client_id
+    if code_input_timeout is None:
+        code_input_timeout = CODE_INPUT_TIMEOUT
+    state = uuid.uuid4()
+    host = host.strip("/")
+    dagshub_url = urllib.parse.urljoin(host, "login/oauth")
+    link_prompt = f"Go to {dagshub_url}/authorize?state={state}&client_id={client_id} and paste the code back in here."
+    code_prompt = "Code:"
+    if code_input_timeout <= 0:
```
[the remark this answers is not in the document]
kbolashev: Using 0/-1 as an infinite timeout is a lot more common + `None` imo means "Unset" not actually None, at least how I often see it used in arguments.
I might change my opinion tomorrow 😄

dagshub/auth/tokens.py
```diff
@@ -0,0 +1,126 @@
...
+
+logger = logging.getLogger(__name__)
+
+APP_TOKEN_TYPE = "app-token"
+
+
+class TokenStorage:
+    def __init__(self, cache_location: str = None, **kwargs):
+        if cache_location is None:
+            cache_location = config.cache_location
```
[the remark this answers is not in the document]
kbolashev: Thanks for the reminder, will change

dagshub/auth/tokens.py
kbolashev: There is no invalidation logic right now.
There isn't even any client-side logic to delete a token if it's actually invalid. This is something that'd need to be added.
Unless it is added I don't see difference in between getting the first or the last token

dagshub/auth/tokens.py
kbolashev: What's the solution? Re-cache the file every time before saving and adding the difference?

dagshub/auth/tokens.py
simonlsk: I meant a user token not an oath token, which BTW I don't think we should cache anyway.

dagshub/common/cli.py
kbolashev: Agree that `dagshub login` is probably better than `dagshub auth login`
Right now the CLI is way too small to care about making a "sensible" tree out of it in terms of verbs, so doing it the flat way is maybe a better idea.

re: passing secrets to the program, I actually completely forgot, but I also wanted to do an env var that will override the whole token logic, and `get_token()` would just return that.
From my chats with Guy, he said he wants there to be basically two flows:
- User just wants to try out DagsHub - let them use OAuth to auth as quickly as possible
- Advanced users should create a token and use that to auth in the client.
I don't really see a point in also keeping username/password auth tbh.

dagshub/auth/tokens.py
simonlsk: I don't really think a solution is necessary rn, if that become a problem we'll think of one...

simonlsk:
Oh, and you need to handle the conflicts - **pay attention** Ido removed all the oath calls in his code and merged, meaning merging will probably remove part of your code.

kbolashev:
> Oh, and you need to handle the conflicts - **pay attention** Ido removed all the oath code and merged, meaning merging will probably remove part of your code.

Will rebase on alpha before merging then, okie

dagshub/auth/oauth.py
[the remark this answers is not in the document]
simonlsk: Fair enough

dagshub/common/cli.py
```diff
@@ -39,34 +39,16 @@ def mount(ctx, **kwargs):
...
-    Add a short-lived OAuth token by logging in at DagsHub
-    """
-    dagshub.auth.add_oauth_token(ctx.obj["host"])
-    print("Successfully stored OAuth token in cache")
+def login(ctx, **kwargs):
+    host = kwargs["host"] or ctx.obj["host"]
+    if kwargs["token"] is not None:
+        dagshub.auth.add_app_token(kwargs["token"], host)
+    else:
+        dagshub.auth.add_oauth_token(host)
```
simonlsk, line 51: Why not straight args here? You're not even passing `**kwargs` to anyone.

simonlsk (approved):

PR #186, opened 2022-10-03
Use token in get default branch

idonov8:


idonov8:
Redundant, @simonlsk is fixing it in a separate PR

PR #188, opened 2022-10-04
Fixing bad cwd handling

kbolashev:
This should fix bugs that happen after you do something like:
```python
os.chdir("<repository root path>")
install_hooks()
os.chdir("data-dir")
os.listdir(".")
```
Before this fix listdir() would return contents of the repo root and not data-dir

### DIFF
--- a/dagshub/streaming/filesystem.py
+++ b/dagshub/streaming/filesystem.py
@@ -80,8 +80,9 @@ class DagsHubFilesystem:
                 if ismount(self.project_root):
                     raise ValueError('No git project found! (stopped at mountpoint {self.project_root})')
                 self.project_root = self.project_root / '..'
+            self.project_root = Path(os.path.abspath(self.project_root))
         else:
-            self.project_root = Path(project_root)
+            self.project_root = Path(os.path.abspath(project_root))
         del project_root
         # TODO: if no Git project found, search for .dvc project?
 
--- a/dagshub/streaming/filesystem.py
+++ b/dagshub/streaming/filesystem.py
@@ -75,12 +75,11 @@ class DagsHubFilesystem:
 
         # Find root directory of Git project
         if not project_root:
-            self.project_root = Path('.')
+            self.project_root = Path(os.path.abspath('.'))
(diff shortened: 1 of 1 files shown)


dagshub/streaming/filesystem.py
```diff
@@ -80,8 +80,9 @@ def __init__(self,
                 if ismount(self.project_root):
                     raise ValueError('No git project found! (stopped at mountpoint {self.project_root})')
                 self.project_root = self.project_root / '..'
+            self.project_root = Path(os.path.abspath(self.project_root))
```
simonlsk, line 83: Why not edit the original `self.project_roo = Path('.')`
And why not keep it relative if it's relative, but relative to cwd?

kbolashev: Re: keeping it relative
Ok, so there's a function called `_relative_path(path)` which is used at the beginning of all functions. 
I assume its "spec" is:
- If the path is inside of the repo, return its path relative to the repo root
- if the path is outside of the repo (among other edgecases) return None
This path is then used to query the api among other things

It does it by calculating the path's relative path to `self.project_root`
IF `self.project_root` itself is relative, then it suddenly becomes cwd-dependent, and whatever path `self.project_root` points to drifts from the actual project root whenever you change directories.

so, _relative_path basically has this behavior (pseudocode snippet, don't assume it will run):
```python
# cwd = git repo root
fs = DagsHubFilesystem()
# fs.project_root = "."
print(fs._relative_path("data")) # Expected and actual: "data"
os.chdir("data")
print(fs._relative_path("nested-data")) # Expected: "data/nested-data", actually returns "nested-data"
```

This is because of [this line](https://github.com/DagsHub/client/blob/alpha/dagshub/streaming/filesystem.py#L163). Since `fs.project_root = "."`, after you chdir to data, the project_root is still `"."`
So when it gets resolved for relative check, it gets resolved to `<project-root>/data`

Got this snippet working in pycharm so you can see it for yourself
image

And with this PR:
image

simonlsk (approved):

PR #189, opened 2022-10-06
Implement chdir

kbolashev:
Don't know how it will work with FUSE.

Current flow:
if dir is outside of repo: cd into it
if dir is inside repo:
- try to cd into it
- if it fails, query list of the dir. If dir lists, then make the dir on the fs and cd into it
- otherwise raise the FileNotFound error you get originally from trying to cd


kbolashev:
[Streaming Client: monkey patch os.chdir()](https://trello.com/c/8PbLOu5l/1375-streaming-client-monkey-patch-oschdir)

PR #191, opened 2022-10-06
Alpha Formatting

simonlsk:


kbolashev (approved):

dagshub/streaming/mount.py
```diff
@@ -20,7 +21,6 @@
               f"Please use install_hooks to access DagsHub hosted files from a python script"
     raise ImportError(err_str)
 
-from fuse import FUSE, FuseOSError, LoggingMixIn, Operations
 
 class DagsHubFUSE(LoggingMixIn, Operations):
     def __init__(self,
```
kbolashev, line 26: We have to put a #noqa here, because importing from fuse checks for existing libfuse and throws an error if there is none installed.
This is why I'm doing this check above

kbolashev (changes requested):
Thanks for doing it before me 😄 
Need to do something with the fuse import (I suggest just noqa-ing)

kbolashev (approved):
Thanks!
I think we should try to handle this one better later at the module import level, just try to catch the particular libfuse exception and shut off if it's not installed.
I might #justdoit that today honestly

PR #194, opened 2022-10-12
Add CLI and oauth to upload

simonlsk:


dagshub/common/cli.py
```diff
@@ -49,5 +54,71 @@ def login(ctx, token, host):
...
+
+def to_log_level(verbosity):
+    if verbosity == 0:
+        return logging.ERROR
+    elif verbosity == 1:
+        return logging.WARN
+    elif verbosity == 2:
+        return logging.INFO
+    elif verbosity >= 3:
+        return logging.DEBUG
```
kbolashev, line 86: Default loglevel for python logging is `logging.WARN`
I feel like hiding warnings behind the verbosity flag is a bad idea and we should do
```suggestion
def to_log_level(verbosity):
    if verbosity == 0:
        return logging.WARN
    elif verbosity == 1:
        return logging.INFO
    elif verbosity >= 2:
        return logging.DEBUG
```

dagshub/common/cli.py
```diff
@@ -49,5 +54,71 @@ def login(ctx, token, host):
...
+@click.argument("repo", callback=validate_repo)
+@click.argument("filename", type=click.Path(exists=True))
+@click.argument("target")
+@click.option("-m", "--message", help="Commit message for the upload")
+@click.option("-b", "--branch", help="Branch to upload the file to")
+@click.option("-u", "--user", callback=validate_user, help="Username and password in the format '<user>:<password>'."
+                                                           "This option is not recommended, instead leave this empty "
+                                                           "to use oauth or specify --token to use an existing "
+                                                           "user token.")
+@click.option("--update", is_flag=True, help="Specify --update to force update an existing file")
```
kbolashev, line 99:
```suggestion
@click.option("--update", is_flag=True, help="Force update an existing file")
```

dagshub/common/logging.py
```diff
@@ -24,13 +24,8 @@ def format(self, record):
 
 
 # create logger with 'spam_application'
-logger = logging.getLogger("DagsHub")
-logger.setLevel(logging.DEBUG)
-
+logger = logging.getLogger()
```
kbolashev, line 27: Code libraries (imo) should never ever fiddle with the root logger, that's gonna ruin some user's day.
Due to how python's logging inheritance works, we can actually do this:
```python
logger = logging.getLogger("dagshub")
```

And as long as every other class in the library gets a logger via 
```python
logger = logging.getLogger(__name__) # say dagshub.streaming.filesystem
```
the logging calls will bubble up to this logger and we should get this format for the whole library.

dagshub/common/cli.py
```diff
@@ -49,5 +54,71 @@ def login(ctx, token, host):
...
+           verbose,
+           update,
+           **kwargs):
+    """
+    Upload FILENAME to REPO at location TARGET.
+    REPO should be of the form <owner>/<repo-name>, i.e nirbarazida/yolov6.
+    TARGET should include the full path inside the repo, including the filename itself.
+    """
+    from dagshub.upload import Repo
+    dagshub.common.logging.logger.setLevel(to_log_level(verbose))
```
kbolashev, line 120: Since this is a CLI, WDYT about doing `logging.BasicConfig()` here with the verbosity level?
That way we/user can debug requests and other used libraries, and not just seeing the logs from our library.

dagshub/common/logging.py
```diff
@@ -24,13 +24,8 @@ def format(self, record):
...
 # create console handler with a higher log level
 ch = logging.StreamHandler()
```
kbolashev, line 29: I like the idea of having a pretty printout.
But at the same time, if the user has a root logger already configured, I think this will end up in all the messages double-printing. Once for the library logger, once for the root logger.

image

kbolashev (changes requested):
The auth code looks good and I don't mind WET here since it's a lot of class state that is used in there
Logging needs some work, but unifying it for the whole library is a great idea, however I'm not sure if it's worth it to try to format it.
Maybe we can do the formatting in the CLI, just because this is supposed to be an entrypoint and we can do whatever we want with the logging there.

Also there are a couple minor linter warnings btw.

dagshub/common/logging.py
simonlsk: We discussed this I think

dagshub/common/cli.py
simonlsk: We discussed this I think

dagshub/common/logging.py
simonlsk: I'll just not temper with this at all, WDYT?

kbolashev (approved):

dagshub/common/logging.py
kbolashev: Good idea, let's just leave formatting for the CLI, and in all other usecases just defer to however the user set up their logging

PR #199, opened 2022-10-18
Change upload argument from commit_message to commit_summary

kbolashev:
This way the user specified message will show up as the first commit line, instead of after "Uploaded files"


simonlsk (changes requested):
Rename back to `commit_message` which is clearer for users, but send it as `commit_summary`.

PR #200, opened 2022-10-18
Handle user trying to add a file to the DataSet more than once

kbolashev:
I wanted to implement it with a `set()` first, but calling `open()` multiple time leads to different descriptors, hence different hashes for the resulting tuple of `(path, file)`

### DIFF
--- a/dagshub/upload/wrapper.py
+++ b/dagshub/upload/wrapper.py
@@ -136,15 +136,18 @@ class Repo:
 
 class DataSet:
     def __init__(self, repo: Repo, directory):
-        self.files = []
+        self.files = {}
         self.repo = repo
         self.directory = directory
         self.request_url = self.repo.get_request_url(directory)
 
     def add(self, file: Union[str, IOBase], path=None):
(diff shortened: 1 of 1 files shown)


dagshub/upload/wrapper.py
```diff
@@ -136,15 +136,17 @@ def full_name(self):
...
         self.request_url = self.repo.get_request_url(directory)
 
     def add(self, file: Union[str, IOBase], path=None):
-        file = self.get_file(file, path)
+        path, file = self.get_file(file, path)
         if file is not None:
-            self.files.append(file)
+            if path in self.files:
+                logger.warning(f"File already staged for upload on path \"{path}\". Overwriting")
+            self.files[path] = file
```
simonlsk, line 149: The item needs to be a tuple of (path, file)

simonlsk (changes requested):
No functional change, adding a warning and preventing network waste, right?

simonlsk (approved):

PR #205, opened 2022-10-23
Feature/create repo client

martintali:


dagshub/upload/wrapper.py
```diff
@@ -24,6 +26,66 @@ def get_default_branch(src_url, owner, reponame, auth):
...
+def create_repo(repo_name, description="", private=False, auto_init=False, gitignores="", license="", readme="", template="notebook-template"):
+    import logging
+    logging.basicConfig(level=logging.DEBUG)
+    import dagshub.auth
+    from dagshub.auth.token_auth import HTTPBearerAuth
+
+    username = config.username
+    password = config.password
+    if username is not None and password is not None:
+        return username, password
```
guysmoilov, line 38: return?

dagshub/upload/wrapper.py
```diff
@@ -24,6 +26,66 @@ def get_default_branch(src_url, owner, reponame, auth):
...
+    userJson = userRes.json()
+    if username is None:
+        username = userJson["username"]
+
+    repoRes = requests.get(urllib.parse.urljoin(config.host, REPO_INFO_URL.format(
+        owner=username,
+        reponame=repo_name
+    )))
+    if repoRes.status_code == HTTPStatus.OK:
+        return Repo(owner=username, name=repo_name, username=username, token=token)
```
guysmoilov, line 56: You don't need any of this:
1. Just let the API return an error if the repo already exists
2. The full repo information, including owner name, gets returned in the JSON result from the create repo API
3. The desired owner might be an organization, as I'll describe in a comment below

dagshub/upload/wrapper.py
```diff
@@ -24,6 +26,66 @@ def get_default_branch(src_url, owner, reponame, auth):
...
+        "private": private,
+        "auto_init": auto_init,
+        "gitignores": gitignores,
+        "license": license,
+        "readme": readme,
+        "project_template": template,
+    }
+
+    res = requests.post(
+        urllib.parse.urljoin(config.host, REPO_CREATE_URL),
```
guysmoilov, line 75: Add an optional parameter for the repo to be created under an org with the url pattern: `POST /org/:org/repos`

dagshub/upload/wrapper.py
```diff
@@ -24,6 +26,66 @@ def get_default_branch(src_url, owner, reponame, auth):
...
+    from dagshub.auth.token_auth import HTTPBearerAuth
+
+    username = config.username
+    password = config.password
+    if username is not None and password is not None:
+        return username, password
+    try:
+        token = dagshub.auth.get_token(code_input_timeout=0)
+    except dagshub.auth.OauthNonInteractiveShellException:
+        logger.debug("Failed to perform OAuth in a non interactive shell")
```
guysmoilov, line 42: I think you shouldn't catch this error, and just silently fail. Let the user code crash.

dagshub/upload/wrapper.py
```diff
@@ -24,6 +26,66 @@ def get_default_branch(src_url, owner, reponame, auth):
...
+
+    res = requests.post(
+        urllib.parse.urljoin(config.host, REPO_CREATE_URL),
+        data,
+        auth=auth)
+
+    if res.status_code != HTTPStatus.CREATED:
+        logger.error(f"Response ({res.status_code}):\n"
+                     f"{res.content}")
+        return
```
guysmoilov, line 82: Raise an error

dagshub/upload/wrapper.py
```diff
@@ -24,6 +26,66 @@ def get_default_branch(src_url, owner, reponame, auth):
...
+    else:
+        logger.debug(f"Response ({res.status_code})\n")
```
guysmoilov, line 84: You know what the status code is - it's 201. No point in logging it

dagshub/upload/wrapper.py
```diff
@@ -24,6 +26,66 @@ def get_default_branch(src_url, owner, reponame, auth):
     return res.json().get('default_branch')
 
 
+def create_repo(repo_name, description="", private=False, auto_init=False, gitignores="", license="", readme="", template="notebook-template"):
+    import logging
+    logging.basicConfig(level=logging.DEBUG)
```
guysmoilov, line 31: Looks like all the logging below should be removed, so no need for a logger anyway

dagshub/upload/wrapper.py
guysmoilov: If we want a `get_or_create` functionality, that can be done with an additional helper function external to this one

guysmoilov (changes requested):

dagshub/upload/wrapper.py
martintali: I'll leave the check for the repo, because the create just fails with no clear message that the repo already exists

dagshub/upload/wrapper.py
```diff
@@ -24,6 +27,57 @@ def get_default_branch(src_url, owner, reponame, auth):
     return res.json().get('default_branch')
 
 
+def create_repo(repo_name, is_org=False, org_name="", description="", private=False, auto_init=False, gitignores="",
+                license="", readme="",
+                template="notebook-template"):
```
guysmoilov, line 32: I do think that the default should be to only have a gitignore, and not a big template with various junk
```suggestion
def create_repo(repo_name, is_org=False, org_name="", description="", private=False, auto_init=False,
                 gitignores="Python", license="", readme="", template=""):
```

dagshub/upload/wrapper.py
```diff
@@ -24,6 +27,57 @@ def get_default_branch(src_url, owner, reponame, auth):
...
+    from dagshub.auth.token_auth import HTTPBearerAuth
+
+    token = dagshub.auth.get_token(code_input_timeout=0)
+    if token is not None:
+        auth = HTTPBearerAuth(token)
+
+    if license is None and readme is None and template is None and gitignores is None:
+        raise RuntimeError(
+            "In order to create the repository please choose a project template. "
+            "Our options are cookiecutter-dagshub-dvc or notebook-template.")
```
guysmoilov, line 43:
```suggestion
            "Creating an empty repository is not supported. "
            "Please choose a valid gitignore, license, readme, or project template. "
            "You can see the list of options here: https://dagshub.com/repo/create . "
            "Our template options are 'cookiecutter-dagshub-dvc' or 'notebook-template' .")
```

dagshub/upload/wrapper.py
```diff
@@ -24,6 +27,57 @@ def get_default_branch(src_url, owner, reponame, auth):
...
+
+    if res.status_code != HTTPStatus.CREATED:
+        logger.error(f"Response ({res.status_code}):\n"
+                     f"{res.content}")
+        if res.status_code == HTTPStatus.UNPROCESSABLE_ENTITY:
+            raise RuntimeError("Repository name is invalid or it already exists.")
+        else:
+            raise RuntimeError("Failed to create the desired repository.")
+    else:
+        logger.debug(f"Response ({res.status_code})\n")
```
guysmoilov, line 75:
```suggestion
```

dagshub/upload/wrapper.py
```diff
@@ -24,6 +27,57 @@ def get_default_branch(src_url, owner, reponame, auth):
...
 
+def create_repo(repo_name, is_org=False, org_name="", description="", private=False, auto_init=False, gitignores="",
+                license="", readme="",
+                template="notebook-template"):
+    import dagshub.auth
+    from dagshub.auth.token_auth import HTTPBearerAuth
+
+    token = dagshub.auth.get_token(code_input_timeout=0)
+    if token is not None:
+        auth = HTTPBearerAuth(token)
```
guysmoilov, line 38: Else - what?
There's no way to proceed, and username and password aren't supported.
It should try to take the username, password, and token from `config`, same as the `Repo` object does.

dagshub/upload/wrapper.py
```diff
@@ -24,6 +27,57 @@ def get_default_branch(src_url, owner, reponame, auth):
...
+                     f"{res.content}")
+        if res.status_code == HTTPStatus.UNPROCESSABLE_ENTITY:
+            raise RuntimeError("Repository name is invalid or it already exists.")
+        else:
+            raise RuntimeError("Failed to create the desired repository.")
+    else:
+        logger.debug(f"Response ({res.status_code})\n")
+
+    repo = res.json()
+    return Repo(owner=repo["owner"]["login"], name=repo["name"], username=repo["owner"]["login"], token=token)
```
guysmoilov, line 78: The username might be an org name - not useful for auth
```suggestion
    return Repo(owner=repo["owner"]["login"], name=repo["name"], token=token)
```

guysmoilov (changes requested):

guysmoilov (approved):
Approved to move forward, but the auth should be improved in a separate PR

dagshub/upload/wrapper.py
```diff
@@ -24,6 +27,56 @@ def get_default_branch(src_url, owner, reponame, auth):
     return res.json().get('default_branch')
 
 
+def create_repo(repo_name, is_org=False, org_name="", description="", private=False, auto_init=False,
+                 gitignores="Python", license="", readme="", template=""):
```
guysmoilov, line 31:
```suggestion
def create_repo(repo_name, is_org=False, org_name="", description="", private=False, auto_init=False,
                gitignores="Python", license="", readme="", template=""):
```

dagshub/upload/wrapper.py
guysmoilov: @martintali They SHOULD supported - I'm just saying that you should be reusing the `config` package here.
(In the current implementation, they aren't checked, only the token - and if there is no token, the call will always fail, so there's no point in even attempting it)

PR #207, opened 2022-10-27
Make custom opener check less eager

kbolashev:
The `file == ""` check fixes an issue preventing FastAI working in Notebooks, especially any display functions.

Basically fastprogress are updating on empty HTML (don't understand reason), which makes Jupyter check if it's a file.
And it does that using `os.stat`, so it tries to do a `os.stat("")`.
We could handle it ourselves, but imo `relative_path("")` returning `"."` is wrong behavior

### DIFF
--- a/dagshub/streaming/filesystem.py
+++ b/dagshub/streaming/filesystem.py
@@ -245,7 +245,7 @@ class DagsHubFilesystem:
         # TODO Include more information in this file
         return b'v0\n'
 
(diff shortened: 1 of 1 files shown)


kbolashev:
[Streaming Client: overeager custom opener check in open](https://trello.com/c/YsT2qJ1y/1433-streaming-client-overeager-custom-opener-check-in-open)

simonlsk (approved):

PR #208, opened 2022-10-27
Feature/create repo client

martintali:


dagshub/upload/wrapper.py
```diff
@@ -32,9 +32,17 @@ def create_repo(repo_name, is_org=False, org_name="", description="", private=Fa
...
 
-    token = config.token or dagshub.auth.get_token(code_input_timeout=0)
-    if token is not None:
-        auth = HTTPBearerAuth(token)
+    username = config.username
+    password = config.password
+    if username is not None and password is not None:
+        auth = username, password
+    else:
+        token = config.token or dagshub.auth.get_token(code_input_timeout=0)
```
guysmoilov, line 40: The timeout override in particular is a mistake in `Repo`, should be fixed there too
```suggestion
        token = config.token or dagshub.auth.get_token()
```

dagshub/upload/wrapper.py
```diff
@@ -32,9 +32,17 @@ def create_repo(repo_name, is_org=False, org_name="", description="", private=Fa
...
+        if token is not None:
+            auth = HTTPBearerAuth(token)
+
+    if auth is None:
+        raise RuntimeError("You can't create a repository without being authenticated.")
```
guysmoilov, line 45: This whole snippet is repeated from the `Repo` code and also according to Kirill the wrapper and maybe the mount.
DRY - this should be refactored to a function in the `dagshub.auth` package which returns a HTTP Auth object.
@idonov8 Please work on refactoring it in a separate PR

guysmoilov (approved):

dagshub/upload/wrapper.py
guysmoilov: `dagshub.streaming.filesystem.DagsHubFilesystem.auth`

PR #209, opened 2022-10-30
Make stat work for remote files

kbolashev:
Before that remote files were only returned after listdir was called at least once.
For now I'm just rerunning listdir in order to update the cache.

Additionally the change to the `dirtree`: now it caches all of the fs entities + their types for the path.
Previously the logic was "if path in dirtree - it's a dir, make a dir and return its stat; if path isn't in it - it's a file, return a dagshub stat", which could lead to returning stats for files that don't exist in the remote at all.


simonlsk (approved):
LGTM

PR #210, opened 2022-10-30
Fixing open() function + Handle binary path arguments

kbolashev:
- Allow it to accept binary paths (allowed in python's docs + used by yolov5)
- Allow writing files inside of repository (+ creates directories if user's trying to write to a dvc-tracked nested dir not on disk)

NOTE: right now this is rebased on branch from #209

### DIFF
--- a/dagshub/streaming/filesystem.py
+++ b/dagshub/streaming/filesystem.py
@@ -288,7 +288,9 @@ class DagsHubFilesystem:
         else:
             return self.__open(file, mode, *args, **kwargs, opener=opener)
 
-    def stat(self, path: PathLike, *, dir_fd=None, follow_symlinks=True):
+    def stat(self, path: Union[str, bytes, PathLike], *, dir_fd=None, follow_symlinks=True):
+        if type(path) is bytes:
+            path = os.fsdecode(path)
         if dir_fd is not None or not follow_symlinks:
             logger.debug("fs.stat - NotImplemented")
             raise NotImplementedError('DagsHub\'s patched stat() does not support dir_fd or follow_symlinks')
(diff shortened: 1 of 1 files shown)


dagshub/streaming/filesystem.py
```diff
@@ -245,7 +246,9 @@ def _special_file(self):
         # TODO Include more information in this file
         return b'v0\n'
 
-    def open(self, file: Union[PathLike, int], mode: str = 'r', *args, opener=None, **kwargs):
+    def open(self, file: Union[bytes, PathLike, int], mode: str = 'r', *args, opener=None, **kwargs):
+        if type(file) is bytes:
+            file = file.decode("utf-8")
```
simonlsk, line 251: Why?

dagshub/streaming/filesystem.py
```diff
@@ -259,17 +262,29 @@ def open(self, file: Union[PathLike, int], mode: str = 'r', *args, opener=None,
...
                     else:
-                        # TODO: After API no longer 500s on FileNotFounds
-                        #       check status code and only return FileNotFound on 404s
-                        raise FileNotFoundError(f'Error finding {relative_path} in repo or on DagsHub')
+                        try:
+                            # Using the fact that stat creates tracked dirs
+                            _ = self.stat(self.project_root / relative_path.parent)
+                        except FileNotFoundError:
+                            raise FileNotFoundError(f"{relative_path.parent} does not exist on the filesystem "
+                                                    f"and is not a tracked directory")
```
simonlsk, line 285: This should raise whatever exception that is thrown by open when `os.path.dirname(filePath)` doesn't exist.

simonlsk (changes requested):

dagshub/streaming/filesystem.py
kbolashev: yolov5 tried to open a file to write with a binary path. Apparently this is an accepted behavior as per docs

simonlsk (approved):

PR #212, opened 2022-10-31
Change monkey-patched functions' signature to be in line with Python's

kbolashev:
`stat`, `listdir` and `scandir` are fine and have the same signature right now. Open's is different and I changed it

### DIFF
--- a/dagshub/streaming/filesystem.py
+++ b/dagshub/streaming/filesystem.py
@@ -230,7 +230,7 @@ class DagsHubFilesystem:
     def __del__(self):
         os.close(self.project_root_fd)
 
-    def _relative_path(self, file: Union[PathLike, int]):
+    def _relative_path(self, file: Union[str, PathLike, int]):
         if isinstance(file, int):
             return None
         if file == "":
@@ -251,7 +251,7 @@ class DagsHubFilesystem:
         # TODO Include more information in this file
         return b'v0\n'
 
-    def open(self, file: Union[bytes, PathLike, int], mode: str = 'r', buffering=-1, encoding=None,
+    def open(self, file, mode='r', buffering=-1, encoding=None,
              errors=None, newline=None, closefd=True, opener=None):
         if type(file) is bytes:
             file = os.fsdecode(file)
@@ -326,7 +326,7 @@ class DagsHubFilesystem:
         filemode = ''.join(filemodes)
         return self.open(path, filemode)
 
-    def stat(self, path: Union[str, bytes, PathLike], *, dir_fd=None, follow_symlinks=True):
+    def stat(self, path, *, dir_fd=None, follow_symlinks=True):
         if type(path) is bytes:
             path = os.fsdecode(path)
         if dir_fd is not None or not follow_symlinks:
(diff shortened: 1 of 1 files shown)


dagshub/streaming/filesystem.py
```diff
@@ -246,7 +246,8 @@ def _special_file(self):
         # TODO Include more information in this file
         return b'v0\n'
 
-    def open(self, file: Union[bytes, PathLike, int], mode: str = 'r', *args, opener=None, **kwargs):
+    def open(self, file: Union[bytes, PathLike, int], mode: str = 'r', buffering=-1, encoding=None,
```
guysmoilov, line 249: I think you need to get rid of the type hints in this and the other hook functions, to ensure compatibility.
Otherwise I don't see any issue but I still need to understand the code better

guysmoilov (changes requested):

kbolashev:
An out of scope addition - added a "monkey-patched" os.open, that is used to actually monkey-patch _pathlib.open.
Turns out on Python<=3.9, `pathlib` uses `os.open` to open stuff, and on 3.10 it uses `io.open`. In 3.11 the whole NormalAccessor construct is completely removed and it always calls os functions instead.

dagshub/streaming/filesystem.py
```diff
@@ -282,12 +293,44 @@ def open(self, file: Union[bytes, PathLike, int], mode: str = 'r', *args, opener
...
+                 (because we lose them when passing to builtin open)
+        WARNING: DO NOT patch actual os.open, because the builtin uses os.open.
+                 This is only for the purposes of patching pathlib.open
+        """
+        if dir_fd is not None:
+            logger.debug("fs.os_open - NotImplemented")
+            raise NotImplementedError('DagsHub\'s patched os.open() (for pathlib only) does not support dir_fd')
+        try:
+            open_mode = "r"
+            # Write modes - calling in append mode, so we create the folders if file doesn't exist, but not truncate it
```
guysmoilov, line 324: Should be better explained like you explained to me verbally - to create any intermediate folders IFF the file exists in the repo

guysmoilov (approved):
I guess I only have readability and documentation comments

dagshub/streaming/filesystem.py
```diff
@@ -282,12 +293,44 @@ def open(self, file: Union[bytes, PathLike, int], mode: str = 'r', *args, opener
...
                             raise err
-                        return self.__open(relative_path, mode, opener=project_root_opener)
+                        # Try to download the file if we're in append modes
+                        if "a" in mode or "+" in mode:
+                            resp = self._api_download_file_git(relative_path)
+                            if resp.ok:
+                                with self.__open(relative_path, 'wb', opener=project_root_opener) as output:
+                                    output.write(resp.content)
+                        return self.__open(relative_path, mode, buffering, encoding, errors, newline,
+                                           closefd, opener=project_root_opener)
```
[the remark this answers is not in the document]
kbolashev: The stat call on dir is, in the end, basically there to create the intermediate folders.
Problem is that if the folders don't exist we can't be sure whether the folder actually doesn't exist, or it's a dvc-tracked folder in the repository and it's just not on the PC.
The stat call on the dir calls the api to see if the folder exists, and then creates all of the intermediaries. If the folder isn't in the repo, then it throws an error

PR #215, opened 2022-11-02
Clean directory name for dataset

idonov8:
I tested names for paths on ` ds.add(file=f, path="./weird_name.exe/")` and it works fine.

### DIFF
--- a/dagshub/upload/wrapper.py
+++ b/dagshub/upload/wrapper.py
@@ -18,7 +18,6 @@ ORG_REPO_CREATE_URL = "api/v1/org/{orgname}/repos"
 USER_INFO_URL = "api/v1/user"
 logger = logging.getLogger(__name__)
 
-
 def get_default_branch(src_url, owner, reponame, auth):
     res = requests.get(urllib.parse.urljoin(src_url, REPO_INFO_URL.format(
         owner=owner,
@@ -84,9 +83,6 @@ def create_repo(repo_name, is_org=False, org_name="", description="", private=Fa
     repo = res.json()
     return Repo(owner=repo["owner"]["login"], name=repo["name"], token=token, branch="main")
 
-def clean_directory_name(directory: str):
-    return os.path.normpath(directory)
-
 class Repo:
     def __init__(self, owner, name, username=None, password=None, token=None, branch=None):
         self.owner = owner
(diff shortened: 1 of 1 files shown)


dagshub/upload/wrapper.py
```diff
@@ -153,7 +155,7 @@ def _log_upload_details(self, data, res, files):
             logger.debug(f"Response ({res.status_code})\n")
 
         if res.status_code == 200:
-            logger.info("Upload finished successfully!")
+            print("Upload finished successfully!")
```
guysmoilov, line 158: Any reason for this?
It's rude of libraries to be directly `print`ing - you might screw up a user's intended output (think of programs that pipe their outputs to other programs, it has to be very precise

dagshub/upload/wrapper.py
```diff
@@ -84,6 +84,8 @@ def create_repo(repo_name, is_org=False, org_name="", description="", private=Fa
     repo = res.json()
     return Repo(owner=repo["owner"]["login"], name=repo["name"], token=token, branch="main")
 
+def clean_directory_name(directory: str):
```
kbolashev, line 87: Make it a static function in DataSet, also probably with a _ so it's "private", no reason for it to be in the top level of the module

kbolashev (changes requested):
linter issues btw

kbolashev (approved):
lgtm 👍

dagshub/upload/wrapper.py
```diff
@@ -130,6 +130,7 @@ def upload_files(self,
 
         if force:
             data["last_commit"] = self._get_last_commit()
+        logger.warning(f'Uploading {len(files)} files to "{self.full_name}"...')
```
simonlsk, line 133: Why is this a warning?

dagshub/upload/wrapper.py
```diff
@@ -153,7 +154,7 @@ def _log_upload_details(self, data, res, files):
             logger.debug(f"Response ({res.status_code})\n")
 
         if res.status_code == 200:
-            logger.info("Upload finished successfully!")
+            logger.warning("Upload finished successfully!")
```
simonlsk, line 157: And here?
@guysmoilov  I saw your comment but I don't think messages like these should be reported as warnings.

PR #217, opened 2022-11-02
Add auth arg to the get_default_branch call

kbolashev:
Allows to finally get the default branch on the private repos


dagshub/upload/wrapper.py
```diff
@@ -152,16 +152,10 @@ def auth(self):
...
-        if username is not None and password is not None:
-            return username, password
-        try:
-            token = self.token or dagshub.auth.get_token(code_input_timeout=0)
-        except dagshub.auth.OauthNonInteractiveShellException:
-            logger.debug("Failed to perform OAuth in a non interactive shell")
-        if token is not None:
-            return HTTPBearerAuth(token)
+        if self.username is not None and self.password is not None:
+            return self.username, self.password
```
idonov8, line 156:
```suggestion
            return HTTPBasicAuth(self.username, self.password)
```

idonov8 (changes requested):

dagshub/upload/wrapper.py
idonov8: for readability

idonov8 (approved):
Resolve conflicts and merge

PR #231, opened 2022-11-06
Send user agent header on all upload requests

idonov8:


kbolashev (approved):
In general the other functions aren't DDA functions, but to keep consistent maybe add it also to the calls in the [filesystem file](https://github.com/DagsHub/client/blob/4294d6200b00e380a577078047e44f3d9520876d/dagshub/streaming/filesystem.py)

btw, maybe at some point we need to start using a [session](https://requests.readthedocs.io/en/latest/user/advanced/#session-objects) for this

idonov8:
I liked the idea of using `Session()`.
In the [filesystem file](https://github.com/DagsHub/client/blob/4294d6200b00e380a577078047e44f3d9520876d/dagshub/streaming/filesystem.py) the headers are already sent, but we should switch to using `Session()` there too in the future.

PR #232, opened 2022-11-07
Better version handling

kbolashev:
1) Moved `__version__` to top level import so you can get it by doing `from dagshub import __version__`
2) Added code to setup.py to import version (as per the [Packaging Guide](https://packaging.python.org/en/latest/guides/single-sourcing-package-version/))


dagshub/__init__.py
```diff
@@ -1,3 +1,4 @@
+__version__ = "0.2.1"
```
idonov8, line 1: Do we need to update this line manually on every release?

idonov8: was it like that before?

kbolashev: Yep, we need to update it on every release.
Before that it was 2 places where it needed to be updated (version.py + in setup.py), now it will be only one place

idonov8 (approved):
OK, I approve. Would be really cool to have the GitHub action change it automatically when we release.

PR #234, opened 2022-11-09
Streaming: Add */site-packages/* to the passthrough path filter

kbolashev:
This should improve performance in use cases where the user has a venv inside of the project repository

Trying to look for modules inside of the project root takes a lot of time because of the fallback request to the server on a nonexistent file.

At the same time, I expect that a possibility of a user:
- having a site-packages folder in their repo/storing modules in a repository
- much more, having them in dvc and not in git
To be close to nil

### DIFF
--- a/dagshub/streaming/filesystem.py
+++ b/dagshub/streaming/filesystem.py
@@ -245,8 +245,12 @@ class DagsHubFilesystem:
         except ValueError:
             return None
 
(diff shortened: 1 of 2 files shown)


TalMalka123 (commented):
it looks good to me but i suggest adding some test to make sure it works and that it doesn't brake anything else

PR #235, opened 2022-11-10
Set default branch for streaming

idonov8:
Add fallback to default branch in case it wasn't specified and the script isn't running from a git repository.
This adds the functionality without changing much.
In general, I think we should avoid getting any information at all from the local `.git` if the remote doesn't match the user specified one.


dagshub/streaming/filesystem.py
```diff
@@ -153,17 +153,23 @@ def _current_revision(self) -> str:
...
-                    raise RuntimeError(f"Current HEAD ({head}) doesn't exist on the remote. "
-                                       f"Please push your changes to the remote or checkout a tracked branch.")
+                    # contents of HEAD is the revision - check that this commit exists on remote
+                    if self.is_commit_on_remote(head):
+                        return head
+                    else:
+                        raise RuntimeError(f"Current HEAD ({head}) doesn't exist on the remote. "
+                                           f"Please push your changes to the remote or checkout a tracked branch.")
+            except FileNotFoundError:
+                logger.warn("No branch was specified, fetching default branch...")
```
kbolashev, line 169:
```suggestion
                logger.debug("Couldn't get branch info from git repository, fetching default branch from the remote...")
```

Debug because I don't see this being relevant to the user that much

dagshub/streaming/filesystem.py
```diff
@@ -153,17 +153,23 @@ def _current_revision(self) -> str:
...
+                owner, reponame = self.parsed_repo_url.path.split("/")[1:]
+                branch = helpers.get_default_branch(owner, reponame, self.auth)
+                logger.warn(f'Set default branch: "{branch}"')
```
kbolashev, line 172:
```suggestion
                logger.debug(f'Set default branch: "{branch}"')
```

dagshub/common/helpers.py
```diff
@@ -0,0 +1,10 @@
+from dagshub.common import config
+import urllib
+import requests
+
+def get_default_branch(owner, reponame, auth):
+    res = requests.get(urllib.parse.urljoin(config.host, config.REPO_INFO_URL.format(
+        owner=owner,
```
kbolashev, line 7:
```suggestion
def get_default_branch(owner, reponame, auth, config=config.host):
    res = requests.get(urllib.parse.urljoin(host, config.REPO_INFO_URL.format(
        owner=owner,
```
We need to keep the ability of the host being overrideable.
I don't see a host arg in any of the users, but still

kbolashev (changes requested):
LGTM but needs minor adjustments.
Also
1) Linter errors
2) Please add the new logic to the docstring of _current_revision (filesystem:L146)

kbolashev (approved):

PR #237, opened 2022-11-16
DagsHubFilesystem: Move from requests to HTTPX

kbolashev:
This should improve the performance of streaming significantly.
Sadly have to use it without the Client object (Benchmark # 4), because it breaks on some multiprocessing usecases, specifically I noticed FastAI with workers had issues.

Benchmark data comparing to Requests, while querying n files from DagsHub (Average of 5 runs over office Wi-Fi).

```
Requests
	1:	1.80s
		Max:2.68s Min: 1.53s
	10:	12.09s
		Max:13.03s Min: 11.16s
	25:	32.60s
		Max:44.42s Min: 27.29s
HTTPX
	1:	0.57s
		Max:0.61s Min: 0.54s
	10:	3.95s
		Max:4.44s Min: 3.78s
	25:	9.16s
		Max:10.02s Min: 8.70s
HTTPX - HTTP2 (Caveat - load balancer communicates with backend in HTTP1.1)
	1:	0.66s
		Max:1.21s Min: 0.47s
	10:	4.00s
		Max:4.68s Min: 3.51s
	25:	8.97s
		Max:9.54s Min: 8.72s
HTTPX - no session
	1:	0.57s
		Max:0.64s Min: 0.54s
	10:	6.00s
		Max:6.16s Min: 5.74s
	25:	15.02s
		Max:15.21s Min: 14.78s
```

### DIFF
--- a/dagshub/streaming/filesystem.py
+++ b/dagshub/streaming/filesystem.py
@@ -85,7 +85,7 @@ class DagsHubFilesystem:
                  'password',
                  'dagshub_remotes',
                  'token',
-                 'http_client',
+                 # 'http_client',
                  '__weakref__')
 
     def __init__(self,
@@ -134,7 +134,7 @@ class DagsHubFilesystem:
         self.password = password or config.password
         self.token = token or config.token
 
-        self.http_client = httpx.Client(auth=self.auth, timeout=None, follow_redirects=True)
+        # self.http_client = httpx.Client(auth=self.auth, timeout=None, follow_redirects=True)
 
         response = self._api_listdir('')
         if response.status_code < 400:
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -40,7 +40,9 @@ def get_version(rel_path: str) -> str:
         "fusepy>=3",
         "appdirs>=1.4.4",
         "pytimedinput>=2.0.1",
-        "click>=8.0.4"],
+        "click>=8.0.4",
```
idonov8, line 43: What's `click`?

dagshub/streaming/filesystem.py
```diff
@@ -494,12 +494,15 @@ def scandir(self, path='.'):
...
-                            headers=config.requests_headers)
+        return httpx.get(f'{self.content_api_url}/{path}',
+                         auth=self.auth, timeout=None, follow_redirects=True,
+                         params={'include_size': 'true'} if include_size else {},
+                         headers=config.requests_headers)
 
     def _api_download_file_git(self, path: str):
-        return requests.get(f'{self.raw_api_url}/{path}', auth=self.auth, headers=config.requests_headers)
+        return httpx.get(f'{self.raw_api_url}/{path}',
+                         auth=self.auth, timeout=None, follow_redirects=True,
```
idonov8, line 504: Make it a const. Maybe a better idea is to use a session with `httpx.Client()` and not repeat it every time :) 
```suggestion
                         auth=self.auth, timeout=TIMEOUT, follow_redirects=True,
```

idonov8 (changes requested):

setup.py
kbolashev: [Package we use for the CLI](https://github.com/pallets/click)

dagshub/streaming/filesystem.py
kbolashev: Can't use the Client object, because it breaks in multiprocessing, will figure something out with the timeout though

PR #246, opened 2022-12-06
1483 - add create repo to client (for new onboarding), fix bug in upload

sdafni:


dagshub/common/cli.py
```diff
@@ -1,10 +1,19 @@
...
 import sys
+from http import HTTPStatus
+
 import click
 
 import dagshub.auth
 from dagshub.common import config
 import dagshub.common.logging
 import logging
+import git
```
kbolashev, line 11: We don't have GitPython as a dependent package, the user-installed dependencies are listed in [`setup.py`](https://github.com/DagsHub/client/blob/master/setup.py#L37).
You should add it in there.
Keep in mind when setting the version that we support Python 3.7 and up

dagshub/common/cli.py
```diff
@@ -105,8 +114,77 @@ def upload(ctx,
...
-    repo.upload(file=filename, path=target, commit_message=message, force=update)
+    try:
+        repo.upload(file=filename, path=target, commit_message=message, force=update)
+    except RuntimeError as re:
+        # if uploaded filename already exists but update is False, a strange error message will appear
+        # since gogs tries to check last commit in this case
+        if "invalid last_commit" in str(re):
+            raise RuntimeError('Verify that the uploaded file is new') from re
+        else:
+            raise re
```
kbolashev, line 125: I agree that the error message about invalid last_commit is bad, but this is not the place to handle it.
We should either fix it in the backend, or wrap the error in `repo.upload()`, so not only CLI gets a better error

dagshub/common/cli.py
```diff
@@ -105,8 +114,77 @@ def upload(ctx,
...
+
+
+@cli.command()
+@click.argument("repo", callback=validate_repo)
+@click.option("-u", "--upload-data", help="Upload data from specified url to new repository")
+@click.option("-c", "--clone", is_flag=True,  help="Clone repository locally")
+@click.option("-v", "--verbose", default=0, count=True, help="Verbosity level")
+def repo_create(
```
kbolashev, line 133: Use [@cli.group()](https://click.palletsprojects.com/en/8.1.x/commands/#merging-multi-commands) so we can have `dagshub repo create`.
We might extend it in the future to be `dagshub repo delete` or something else

dagshub/common/cli.py
```diff
@@ -105,8 +114,77 @@ def upload(ctx,
...
+                repo,
+                upload_data,
+                clone,
+                verbose):
+    """
+    repo_create
+    example:  dagshub repo-create sdafni.yd/yuvtutorial14 -u "http://0.0.0.0:8080/index.html" --checkout
+    """
```
kbolashev, line 141:
```suggestion
    repo_create
    example:  dagshub repo-create sdafni.yd/yuvtutorial14 -u "http://0.0.0.0:8080/index.html" --clone
    """
```

dagshub/common/cli.py
```diff
@@ -105,8 +114,77 @@ def upload(ctx,
...
+    logger = logging.getLogger()
+    logger.setLevel(to_log_level(verbose))
+
+    owner, repo_name = repo
+    INITIAL_DATA_FILE_NAME = 'repo_created_initial_data'
```
kbolashev, line 146: No need to ask the user for their username if they can only create the repo for themselves or an organization they're in.
You can actually get the full name of the repo if you use [`repo.full_name`](https://github.com/DagsHub/client/blob/master/dagshub/upload/wrapper.py#L332)

There is the edge case of organizations, but that would require another flag

dagshub/common/cli.py
```diff
@@ -105,8 +114,77 @@ def upload(ctx,
...
+
+    #
+    # create the repo in dagshub
+    #
+    repo = create_repo(repo_name)
+    logger.info(f"Created repo: {DEFAULT_HOST}/{owner}/{repo_name}.git")
```
kbolashev, line 152: We need to have an ability for users to override the host and not only do it with dagshub.com
You can get it from the cli's context `ctx` (need to specify the ctx argument and add a decorator. You can look at the other functions to see how they do it)

dagshub/common/cli.py
```diff
@@ -105,8 +114,77 @@ def upload(ctx,
...
+        #
+        # get data file,
+        # if repo was cloned - copy the file to local location, commit and push to repo.
+        # if not - upload using gogs api, and delete the local copy
+        #
+        res = http_request("GET", upload_data)
+        if res.status_code != HTTPStatus.OK:
+            raise RuntimeError(f"Could not get file from source (response: {res.status_code}), repo created")
+
+        with open(INITIAL_DATA_FILE_NAME, 'w+') as fh:
```
kbolashev, line 171:
```suggestion
        with open(INITIAL_DATA_FILE_NAME, 'w') as fh:
```
we need to overwrite this file, otherwise the user can upload corrupted data.

Another note: the dataset can be an archive. If the user supplies an archive we need to unpack it.
It's fine if we don't support all formats, but has some ways to do [tars](https://docs.python.org/3/library/tarfile.html) and [zips](https://docs.python.org/3/library/zipfile.html), which should be enough if we warn the user about that.

The MNIST dataset is just gzipped, and archived in a specific way specifically for ML, so we don't need to do anything for the MVP, but it's still something I'd expect the client can handle

dagshub/common/cli.py
```diff
@@ -105,8 +114,77 @@ def upload(ctx,
...
+            fh.write(res.text)
+
+        if clone:
+            os.rename(INITIAL_DATA_FILE_NAME, f"{repo_name}/{INITIAL_DATA_FILE_NAME}")
+            local_repo = git.Repo(repo_name)
+            local_repo.git.add('--all')
+            local_repo.git.commit('-m', DEFAULT_COMMIT_MESSAGE)
+            origin = local_repo.remote(name='origin')
+            origin.push()
+            logger.info(f"Data file named {INITIAL_DATA_FILE_NAME} committed and pushed to repo")
```
kbolashev, line 181: The idea is to actually upload the dataset as a DVC folder, because it's data and not code.
You can either use the [`create_dataset`](https://github.com/DagsHub/client/blob/master/dagshub/upload/wrapper.py#L29) function from the wrapper, which would require some shuffling in this function, or repeat the steps that create_dataset.

dagshub/common/cli.py
```diff
@@ -105,8 +114,77 @@ def upload(ctx,
...
+            repo.upload(file=INITIAL_DATA_FILE_NAME)
+            os.remove(INITIAL_DATA_FILE_NAME)
+            logger.info(f"Data file named {INITIAL_DATA_FILE_NAME} uploaded to repo")
+
+
+
 
 
-if __name__ == "__main__":
-    cli()
```
kbolashev, line 112: Why delete this?

dagshub/upload/wrapper.py
```diff
@@ -341,6 +341,27 @@ def full_name(self):
         return f"{self.owner}/{self.name}"
 
 
+
+    def _get_last_commit(self):
+        """
+        The _get_last_commit function returns the last commit sha for a given branch.
+        It is used to check if there are any new commits in the repo since we last ran our dag.
```
kbolashev, line 348: Great catch! Didn't notice that this function isn't actually in the object it gets called at 🙃

kbolashev (changes requested):
For some reason tests aren't running, this is strange.... 🤔

dagshub/common/cli.py
sdafni: done

dagshub/common/cli.py
sdafni: done (pending push)

dagshub/common/cli.py
sdafni: done (pending push)

dagshub/common/cli.py
sdafni: **it can be done easily, but the help looks wiered when i do it:** 

`dagshub      
Usage: dagshub [OPTIONS] COMMAND [ARGS]...

Options:
  --host TEXT  Hostname of DagsHub instance
  --help       Show this message and exit.

Commands:
  login   Initiate an Oauth authentication process.
  mount   Mount a DagsHub Storage folder via FUSE
  repo    commands on a repo: create
  upload  Upload FILENAME to REPO at location TARGET.

(venv_test) yuvaldafni@h-MacBook-Pro-sl-yuval yuv_tutorial % dagshub repo
Usage: dagshub repo [OPTIONS] COMMAND [ARGS]...

  commands on a repo: create

Options:
  --help  Show this message and exit.

Commands:
  create  create a repo, example: dagshub repo create mytutorial -u...
`

so i prefer to do it only when we realy need more than one command

kbolashev: You can do custom help with a docstring
```
@cli.group()
def repo():
    """Working with DagsHub repositories"""

```

sdafni: done (pending push)

dagshub/common/cli.py
sdafni: done (pending push)

dagshub/common/cli.py
sdafni: done (pending push)

dagshub/common/cli.py
sdafni: done

dagshub/common/cli.py
sdafni: done

dagshub/upload/wrapper.py
```diff
@@ -26,19 +27,28 @@
...
     :return : Repo object of the repository created
     """
-    repo = create_repo(repo_name, org_name=org_name, private=private)
-    dir = repo.directory(repo_name)
+    if not repo:
+        repo = create_repo(repo_name, org_name=org_name, private=private)
+
+    dir = repo.directory(data_dir)
     dir.add_dir(local_path, glob_exclude)
     return repo
```
kbolashev, line 53: This changes the default behavior of `create_dataset` when no repo is passed. Now it expects the data to always be in the `data_dir`, instead of the repo name.

Maybe spin your behavior into an `add_dataset` function on the Repo object, because the logic here is already getting a bit confusing

dagshub/common/cli.py
```diff
@@ -114,77 +115,97 @@ def upload(ctx,
...
+    new_data_dir = DEFAULT_DATA_DIR_NAME
 
-    #
-    # create the repo in dagshub
-    #
-    repo = create_repo(repo_name)
-    logger.info(f"Created repo: {DEFAULT_HOST}/{owner}/{repo_name}.git")
+    # clean tmp dir if exists from previous run
+    if os.path.exists(new_data_dir):
+        shutil.rmtree(new_data_dir)
```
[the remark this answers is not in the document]
sdafni: i did it with a uuid named dir

dagshub/common/cli.py
```diff
@@ -138,71 +139,67 @@ def create(ctx,
...
            verbose):
     """
-    create a repo and upload data
-    example 1:  dagshub repo-create mytutorial -u "http://example.com/data.csv" --clone
-
+    create a repo and:\n
+    optional- upload files to 'data' dir,
+     .zip and .tar files are extracted, other formats copied as is.
+    optional- clone repo locally.\n
+    example 1:  dagshub repo-create mytutorial -u "http://example.com/data.csv" --clone\n
```
kbolashev, line 146:
```suggestion
    create a repo and:
    optional- upload files to 'data' dir,
     .zip and .tar files are extracted, other formats copied as is.
    optional- clone repo locally.
    example 1:  dagshub repo-create mytutorial -u "http://example.com/data.csv" --clone
```

kbolashev (approved):
LGTM 👍

PR #254, opened 2022-12-30
Added Repo Initialization

jinensetpal:


dagshub/common/cli.py
```diff
@@ -52,6 +55,106 @@ def mount(ctx, verbose, **kwargs):
...
+
+
+def _init(repo_name=None, repo_owner=None, url=None, root=None,
+          host=config.DEFAULT_HOST, mlflow=True, dvc=False):
+    from ..upload.wrapper import create_repo
+    import urllib
+
+    # Setup required variables
+    root = root or get_project_root(Path(os.path.abspath('.')))
+
```
kbolashev, line 77: What happens if the user calls that outside of a repository while not specifying the root argument?
I see that you're throwing a ValueError with a good enough error message, but it'll be better if you print out that the user needs to do this in a repo, instead of scaring them with a stack trace

dagshub/common/cli.py
```diff
@@ -52,6 +55,106 @@ def mount(ctx, verbose, **kwargs):
...
+    if not url:
+        if repo_name is not None and repo_owner is not None:
+            url = urllib.parse.urljoin(host, f'{repo_owner}/{repo_name}')
+        else:
+            with open(root / '.git' / 'config', 'r') as cfg:
+                for line in cfg.readlines():
+                    if host in line:
+                        url = line.split()[-1][:-4]
+                        break
```
kbolashev, line 89: You don't have to change this, but we have [GitPython](https://gitpython.readthedocs.io/en/stable/) as a dependency now, so we can ask it to handle remotes for us in a less fragile way:
https://stackoverflow.com/a/60606447

dagshub/common/cli.py
```diff
@@ -52,6 +55,106 @@ def mount(ctx, verbose, **kwargs):
...
+    elif url[-4] == '.':
+        url = url[:-4]
+
+    if not (repo_name and repo_owner):
+        splitter = lambda x: (x[-1], x[-2]) # noqa E721
+        repo_name, repo_owner = splitter(url.split('/'))
+
+    if None in [repo_name, repo_owner, url]:
+        raise ValueError('Host not found; please specify the remote url with --url')
```
kbolashev, line 98: This error message doesn't actually reflect what's happening. The user might have specified the --url and put in a wrong URL, and will be confused.
"Host not found" also doesn't reflect what's actually happening here.
Suggestion:
```suggestion
        raise ValueError('Could not parse repository owner and name. Make sure you specify either a link to the repository with --url or a pair of --repo-owner and --repo-name').
```

Also the thing about not scaring the user with stack traces still applies.

dagshub/common/cli.py
```diff
@@ -52,6 +55,106 @@ def mount(ctx, verbose, **kwargs):
...
+        owner=repo_owner,
+        reponame=repo_name)), auth=bearer or auth)
+    if res.status_code == 404:
+        create_repo(repo_name)
+
+    # Configure MLFlow
+    if mlflow:
+        os.environ['MLFLOW_TRACKING_URI'] = f'{url}.mlflow'
+        os.environ['MLFLOW_TRACKING_USERNAME'] = auth[0]
+        os.environ['MLFLOW_TRACKING_PASSWORD'] = auth[1]
```
kbolashev, line 122: This will not persist for the console session and will only work for the duration of the python script.
I talked about it with Nir when were considering doing a `dagshub init` as a CLI command, and that was one of my concerns.
It's preeeetty hard to export env vars from code like that. You can look at what conda and venv are doing, but this is not a path I think we want to go down, and instead keep it leaner and have it as an init command that the user calls IN the python script before running their own code. [WandB integration with PyCaret is a good example](https://wandb.ai/cayush/customer_churn/reports/Predict-Customer-Churn-using-PyCaret-and-W-B--VmlldzoxNzM4MjA5#setup)

kbolashev (changes requested):

dagshub/__init__.py
```diff
@@ -1,7 +1,11 @@
 __version__ = "0.2.7"
 from .logger import DAGsHubLogger, dagshub_logger
+from .auth import get_token as login
```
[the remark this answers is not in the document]
jinensetpal: The goal was to provide users a pythonic interface to `dagshub login`. They should be able to authorize DH without having to save tokens locally.

dagshub/common/config.py
```diff
@@ -32,3 +32,9 @@
 requests_headers = {"user-agent": USER_AGENT + custom_user_agent_suffix}
 http_timeout = os.environ.get(HTTP_TIMEOUT_KEY, 30)
 REPO_INFO_URL = "api/v1/repos/{owner}/{reponame}"
+
+# DVC config templates
+CONFIG_CORE = "[core]\n    autostage = true\n    analytics = false\n"
+CONFIG_REMOTE = "['remote \"{remote}\"']\n    url = {url}.dvc\n"
+CONFIG_LOCAL = "['remote \"{remote}\"']\n    auth = basic\n    user = {token}\n    password = {token}\n"
+CONFIG_GITIGNORE = "/config.local\n/tmp\n/cache"
```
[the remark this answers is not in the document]
jinensetpal: I tested this, I was able to push to the dvc remote using password:password authentication. Am I missing something?

dagshub/__init__.py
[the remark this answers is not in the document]
jinensetpal: Cool! I agree, I'll remove this.

dagshub/common/config.py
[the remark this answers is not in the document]
kbolashev: We allow logging in with tokens (both as username, and as username:password), but not with passwords, if that works that's really bad and we should fix it 🙃
You sure you weren't using token as a password? 
I will take a closer look

dagshub/common/cli.py
```diff
@@ -52,6 +55,106 @@ def mount(ctx, verbose, **kwargs):
...
+@click.option("--url", help="DagsHub remote url; either provide --url or repo_name and repo_owner")
+@click.option("--host", default=config.DEFAULT_HOST, help="DagsHub instance to which you want to login")
+@click.option("--mlflow", default=True, is_flag=True, help="Set up MLFlow environment with DagsHub remote")
+@click.option("--dvc", default=False, is_flag=True, help="Set up DVC with DagsHub remote")
+@click.pass_context
+def init(ctx, repo_name, repo_owner, url, host, mlflow, dvc):
+    _init(repo_name, repo_owner, url, host, mlflow, dvc)
+
+
+def _init(repo_name=None, repo_owner=None, url=None, root=None,
```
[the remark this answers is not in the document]
jinensetpal: I wrote it this way based on a [SO post](https://stackoverflow.com/a/64181763/10671309) that said click recommends this format.

We need to call the `_init` function pythonically, and exposing the function with the click's overriders is not recommended.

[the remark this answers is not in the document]
jinensetpal: I'll make sure to move the function to somewhere within the common module, per the other comment!

dagshub/common/cli.py
jinensetpal: Should I remove init from the CLI entirely, or keep it for DVC and repo creation?

jinensetpal:
- [x] remove login
- [x] cleanup imports
- [x] remove MLFlow from cli
- [x] better error messages
- [x] move `_init`
- [x] use GitPython
- [x] use ConfigParser

dagshub/common/cli.py
kbolashev: I don't see much point in the init function in the CLI to be honest

So we have three things that it does:
1) Create a repo if it doesn't exist on dagshub
2) Init dvc installation
3) MLFlow

And for these we have basically:
1) This one is already served by the create repo function. It can even clone the repo after creation
2) Already served by dvc init, but I can see at least some value there in terms of setting up the dvc remote
3) Doesn't work

So the only thing I'd really expose in the CLI is dvc init, but with maybe a different name. Maybe something like `dagshub dvc set-remote`, but this is up for you to decide

dagshub/common/helpers.py
```diff
@@ -31,3 +40,99 @@ def http_request(method, url, **kwargs):
...
+            raise ValueError('No git project found! (stopped at mountpoint {root}). \
+                             Please run this command in a git repository.')
+        root = root / '..'
+    return Path(root)
+
+
+def _init(repo_name=None, repo_owner=None, url=None, root=None,
+          host=config.DEFAULT_HOST, mlflow=True, dvc=False):
+    # Setup required variables
+    root = root or get_project_root(Path(os.path.abspath('.')))
```
kbolashev, line 57: Wrap in a try-catch and show the error message instead of a scary stack trace.

Also if root is specified check that it is actually a git repo

dagshub/common/helpers.py
```diff
@@ -31,3 +40,99 @@ def http_request(method, url, **kwargs):
...
+        write = True
+
+        dvc_config = configparser.ConfigParser()
+        dvc_config_local = configparser.ConfigParser()
+        dvc_config.read(root / '.dvc' / 'config')
+        dvc_config_local.read(root / '.dvc' / 'config.local')
+
+        if 'core' not in dvc_config.sections():
+            dvc_config['core'] = {'analytics': False,
+                                  'autostage': True}
```
kbolashev, line 116: I think we shouldn't be enforcing our opinion on what dvc features to toggle on the users, autostage specifically is very much a surprising change.

kbolashev (changes requested):
I'm liking this already, just a couple more changes + the CLI discussion up above

dagshub/common/helpers.py
jinensetpal: I'll clear autostage, but I think analytics being active by default is an unethical practice, so I'm in strong support for that to stay. 

Can we keep that, or should I remove it too?

jinensetpal:
- [x] removed core config settings
- [ ] better error handling
- [x] restructured cli for dvc setup

kbolashev (approved):
👍

PR #260, opened 2023-01-04
Fixed Logical Bug with `init`

jinensetpal:
Allowed no git root if dvc is not initialized.

### DIFF
--- a/dagshub/__init__.py
+++ b/dagshub/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.2.9"
+__version__ = "0.2.9.1"
 from .logger import DAGsHubLogger, dagshub_logger
 from .common.helpers import init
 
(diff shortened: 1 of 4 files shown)


dagshub/__init__.py
```diff
@@ -1,4 +1,4 @@
-__version__ = "0.2.9"
+__version__ = "0.2.9.1"
```
kbolashev, line 1: Don't add additional dots. Looked at what we had previously in our releases and it should be like this:
```suggestion
__version__ = "0.2.9v2"
```

kbolashev (approved):
Change the version number and feel free to merge

dagshub/__init__.py
kbolashev: Also in general we shouldn't be bumping version numbers in PRs because they aren't always merged synchronously

PR #261, opened 2023-01-05
Fix uploads with path separators not working correctly on Windows

kbolashev:
We were using Python's `os.path`, that has os-dependent separators, when we should have used `posixpath`, because the paths sent to the server are POSIX.

### DIFF
--- a/dagshub/upload/wrapper.py
+++ b/dagshub/upload/wrapper.py
@@ -465,7 +465,6 @@ class DataSet:
 
         return posixpath.normpath(directory)
 
-
     @staticmethod
     def get_file(file: Union[str, IOBase], path=None):
         """
--- a/dagshub/upload/wrapper.py
+++ b/dagshub/upload/wrapper.py
@@ -1,4 +1,5 @@
 import json
+import posixpath
 
 import urllib
 import os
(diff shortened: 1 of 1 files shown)


kbolashev:
[DDA Windows issues while uploading nested folders](https://trello.com/c/Z78wqT7f/1544-dda-windows-issues-while-uploading-nested-folders)

dagshub/upload/wrapper.py
```diff
@@ -427,7 +428,7 @@ def add_dir(self, local_path, glob_exclude=""):
         for root, dirs, files in os.walk(local_path):
```
idonov8, line 428: Could you replace `os.walk` here with something with `posixpath`?

idonov8 (approved):
Looks good overall, I'd consider removing the `import os` entirely if possible

dagshub/upload/wrapper.py
kbolashev: Doesn't look like it's possible, but maybe I can cast the path to a posixpath

idonov8: COOL, MERGE IT

kbolashev: nvm I already return a posixpath at that point

PR #262, opened 2023-01-08
Enhancement: improve logging, move it from log.warnings to prints with ability to suppress

kbolashev:


kbolashev:
[PyCaret client modifications](https://trello.com/c/CEcHxdy7/1552-pycaret-client-modifications)

dagshub/common/helpers.py
```diff
@@ -50,85 +48,11 @@ def get_project_root(root):
...
-            with open(root / '.dvc' / '.gitignore', 'w') as config_gitignore:
-                config_gitignore.write(config.CONFIG_GITIGNORE)
-
-    print('Repository initialized!')
+def log_message(msg, logger=None):
+    """
+    Logs message to the info of the logger + prints, unless the printing was suppressed
+    """
+    if not config.quiet:
+        print(msg)
```
idonov8, line 56: Why not `log.warning()`?
I remember @guysmoilov saying a python library shouldn't use `print()`

idonov8 (changes requested):
looks **_Cool_**

dagshub/common/helpers.py
kbolashev: PyCaret are against spamming warning logs with informational messages, those should be kept for warnings.
This right now is basically a compromise where:
- We have the info logs that are actually info(rmational)
- They are also printed out by default
- If you pass DAGSHUB_QUIET you're not getting the prints, but still getting the info logs

Which is tbh the compromise that satisfies most of the requirements.
Guy's complaint about prints was inability to turn them off

kbolashev: @idonov8

idonov8 (approved):
**_COOL 🐶_**

PR #264, opened 2023-01-17
Make the OAuth prompt more noticeable and prettier

kbolashev:
Also we now open the authorization link using `webbrowser.open` which should work everywhere
This PR introduces [rich](https://github.com/Textualize/rich) as a dependency, opening way to prettify other prompts later down the line.

Example of the new prompt (censored in post):
image

Tested platforms:
- Colab - Code Cell, Bash execution using `!`
- Mac - iTerm, IntelliJ Terminal, IntelliJ Run
- Windows - cmd, VSCode Terminal

### DIFF
--- a/dagshub/auth/oauth.py
+++ b/dagshub/auth/oauth.py
@@ -27,7 +27,7 @@ def oauth_flow(
     webbrowser.open(auth_link)
 
     rich_console.print("[bold]:exclamation::exclamation::exclamation: AUTHORIZATION REQUIRED "
-                  ":exclamation::exclamation::exclamation:[/bold]", justify="center")
+                       ":exclamation::exclamation::exclamation:[/bold]", justify="center")
     # Doing raw prints here because the rich syntax breaks in colab
     # Printing them line by line, because the link has to be its own print in order for Colab parser to correctly parse
     # the whole link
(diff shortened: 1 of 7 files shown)


dagshub/auth/oauth.py
```diff
@@ -19,15 +21,25 @@ def oauth_flow(
...
-        f"{dagshub_url}/middleman",
-        data={"request_id": middle_man_request_id}, timeout=None
-    )
+    rich_console.print("[bold]:exclamation::exclamation::exclamation: AUTHORIZATION REQUIRED "
+                  ":exclamation::exclamation::exclamation:[/bold]", justify="center")
+    # Doing raw prints here because the rich syntax breaks in colab
+    # Printing them line by line, because the link has to be its own print in order for Colab parser to correctly parse
+    # the whole link
+    print("\n\nOpen the following link in your browser to authorize the client:")
+    print(auth_link)
```
simonlsk, line 35: Does the rich text let you put colors /

kbolashev: Rich does have colors, but I have to use regular boring prints here, otherwise colab breaks.
I can probably print the first line with rich and the auth link with regular print, let me check a bit later if colab gods would accept this

simonlsk (approved):
* Fix lint errors
* LGTM

Long term we should probably have an intermediate printing module that abstracts away the choice of icons and colors, wiith premade templates.

PR #265, opened 2023-01-22
Changed os.rename() to shutil.move() to fix cloning errors where /tmp is in a separate file system.

jacob-zietek:
On certain Linux flavors like EndeavorOS, `/tmp` is mounted to a separate file system than `/home`. This causes `os.rename()` to throw `OSError: [Errno 18] Invalid cross-device link: 'enron.csv' -> '/tmp/tmpah9csemy/enron.csv'` while doing the Dagshub tutorial. `shutil.move()` handles this edge case.

["If the destination is on the current filesystem, then os.rename() is used. Otherwise, src is copied to dst using copy_function and then removed."](https://docs.python.org/3/library/shutil.html#shutil.move)

### DIFF
--- a/dagshub/common/cli.py
+++ b/dagshub/common/cli.py
@@ -1,6 +1,7 @@
 import os
 import sys
 import tempfile
+import shutil
 
 import click
 import logging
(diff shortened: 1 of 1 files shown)


kbolashev:
Thank you very much for catching this and providing a fix!

kbolashev (approved):

PR #266, opened 2023-01-23
Allow uploading folders via dagshub upload from CLI

kbolashev:
This also allows the regular `repo.upload` function to upload folders and not just files.
Also along the way added a prettier progress bar for uploading the folder, that shows some progress to the user.
And also the commit message for uploading directories/datasets is now a bit better

Tested in Colab, the progress bar doesn't show up there, but the prints with "Uploading x files" still show up, so I consider that ok

### DIFF
--- a/dagshub/upload/wrapper.py
+++ b/dagshub/upload/wrapper.py
@@ -8,7 +8,9 @@ import fnmatch
 from typing import Union
 from io import IOBase
 import httpx
-from dagshub.common import config, helpers
+import rich.progress
+
+from dagshub.common import config, helpers, rich_console
 from http import HTTPStatus
 import dagshub.auth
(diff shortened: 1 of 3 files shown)


dagshub/upload/wrapper.py
```diff
@@ -41,7 +43,7 @@ def create_dataset(repo_name, local_path, glob_exclude="", org_name="", private=
     """
     repo = create_repo(repo_name, org_name=org_name, private=private)
     dir = repo.directory(repo_name)
-    dir.add_dir(local_path, glob_exclude)
+    dir.add_dir(local_path, glob_exclude, commit_message="Initial dataset commit")
```
sdafni, line 46: same string twice, better have a const in wrapper.py

sdafni (approved):
tiny comment

PR #271, opened 2023-02-26
Bug/close files after upload

kbolashev:
If you add files to a dataset upload via `ds.add`, they are not closed after user does `ds.commit` to add them.
This is a problem on windows, where opening any file is exclusive. 
So if a user is uploading "temporary" output to their repo, for example:

```python
with open("out/model-output.txt", "w") as f:
  f.write("Very cool model")

repo = Repo("user", "repo")
ds = repo.directory()
ds.add("out/model-output.txt", "model-output-today.txt")
ds.commit("Added model output")

os.remove("out/model-output.txt")
```

The script crashes on the remove, since the file is still opened by the DataSet.

[Originally raised by Mark in Discord](https://discord.com/channels/698874030052212737/705302775784800278/1074278169672155146)

Caveat: now we close ALL files, even if user explicitly used `ds.add()` to add an IOBase and not a path.
I don't think this is a problem. We don't seek back on the passed IOBase anyway, so it's gonna be seeked to its end. Shouldn't be a big problem that it's closed now.

### DIFF
--- a/dagshub/upload/wrapper.py
+++ b/dagshub/upload/wrapper.py
@@ -5,7 +5,7 @@ import urllib
 import os
 import logging
 import fnmatch
-from typing import Union
+from typing import Union, Tuple, BinaryIO, Dict
 from io import IOBase
 import httpx
 from dagshub.common import config, helpers
@@ -387,7 +387,7 @@ class DataSet:
 
         """
 
-        self.files = {}
+        self.files: Dict[os.PathLike, Tuple[os.PathLike, BinaryIO]] = {}
         self.repo = repo
         self.directory = self._clean_directory_name(directory)
         self.request_url = self.repo.get_request_url(directory)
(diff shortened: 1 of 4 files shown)


dagshub/upload/wrapper.py
```diff
@@ -387,7 +387,7 @@ def __init__(self, repo: Repo, directory: str):
 
         """
 
-        self.files = {}
+        self.files: Dict[os.PathLike, Tuple[os.PathLike, BinaryIO]] = {}
         self.repo = repo
         self.directory = self._clean_directory_name(directory)
         self.request_url = self.repo.get_request_url(directory)
```
sdafni, line 393: not related to your change but self.request_url not used

sdafni (approved):

dagshub/upload/wrapper.py
kbolashev: Thanks for noticing! Guess I'll get rid of it

PR #277, opened 2023-03-27
1634 Prevent users from mounting multiple filesystems in the same location

kbolashev:
# Spec
We want to prevent users from accidentally mounting multiple repositories into the same folder and then getting conflicting data.
On DagshubFilesystem creation I now check a static map of all "mounted" filesystems and if there is one mounted in either of:
- Same path
- Parent path of the path passed to the DagsHubFilesystem
- Child path of the path
I throw a `FilesystemAlreadyMountedError`

The map is cleared of a mounted filesystem on:
- Object cleanup (`__del__`)
- `unmount_hooks()`

# Testing
Added test cases that test most of this behavior (didn't test the object cleanup)

# Docs
Didn't add any :(

### DIFF
--- a/dagshub/streaming/errors.py
+++ b/dagshub/streaming/errors.py
@@ -2,9 +2,11 @@ from pathlib import Path
 
 
 class FilesystemAlreadyMountedError(Exception):
-    def __init__(self, path: Path, revision: str):
+    def __init__(self, path: Path, repo: str, revision: str):
         self.path = path
+        self.repo = repo
         self.revision = revision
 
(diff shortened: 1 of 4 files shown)


kbolashev:
[Fix behavior of multiple DagsHub filesystem objects mounting to the same location](https://trello.com/c/ioUESnef/1634-fix-behavior-of-multiple-dagshub-filesystem-objects-mounting-to-the-same-location)

sdafni (approved):

PR #279, opened 2023-04-02
Handle continuation token in the Storage Content API

kbolashev:
**Only merge once the changes are deployed on the backend!**

# Spec
 
Content API for storages now returns struct with token pagination. This PR handles it.
Additionally, if there is a `next_token` field, we requery the backend to get all of the entries.

# Testing

Added test that handles the pagination case


kbolashev:
[[Python Client] Handle new changes on the storage endpoints](https://trello.com/c/tb99GhSl/1698-python-client-handle-new-changes-on-the-storage-endpoints)

Issue #281, opened 2023-04-17
Improve default behavior for repo.upload


deanp70:
In this code snippet: https://github.com/DagsHub/client/blob/master/dagshub/upload/wrapper.py#L177

The doc string is not up to date.

Also, `remote_path` defaults to `None`, which then fails with a "mysterious" error when run without a remote path provided. We should make it a required parameter or provide better defaults.

krishnaduttPanchagnula:
i have done the following in my PR #285 
- Added a better erroring mechanism to detect the if there are any arguments missing, when running a method
- Added remote_path as required parameter
- Updated the Docstring

PR #285, opened 2023-05-04
made remote_path default and updated docstring

krishnaduttPanchagnula:
This solves the issue number #281

### DIFF
--- a/dagshub/upload/wrapper.py
+++ b/dagshub/upload/wrapper.py
@@ -194,8 +194,8 @@ class Repo:
     def upload(
         self,
         local_path: Union[str, IOBase],
-        commit_message=DEFAULT_COMMIT_MESSAGE,
         remote_path,
+        commit_message=DEFAULT_COMMIT_MESSAGE,
         **kwargs,
     ):
         """
@@ -570,4 +570,3 @@ class DataSet:
             file_list, self.directory, commit_message=commit_message, *args, **kwargs
         )
         self._reset_dataset()
-        
--- a/dagshub/upload/wrapper.py
+++ b/dagshub/upload/wrapper.py
@@ -198,7 +198,7 @@ class Repo:
             dir_to_upload = self.directory(remote_path if remote_path is not None else DEFAULT_REMOTE_PATH)
             dir_to_upload.add_dir(local_path, commit_message=commit_message, **kwargs)
         else:
-            file_to_upload = DataSet.get_file(local_path, remote_path if remote_path is not None else DEFAULT_REMOTE_PATH)
+            file_to_upload = DataSet.get_file(local_path, remote_path)
             self.upload_files([file_to_upload], commit_message=commit_message, **kwargs)
 
     def upload_files(
--- a/dagshub/upload/wrapper.py
+++ b/dagshub/upload/wrapper.py
@@ -9,7 +9,6 @@ from typing import Union, Tuple, BinaryIO, Dict
 from io import IOBase
 import httpx
 import rich.progress
-from functools import wraps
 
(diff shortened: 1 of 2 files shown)


kbolashev:
Hi Krishna!

Thank you very much for the PR!
I understand that you've implemented the PR according to what's been said in the issue, however we try to have unsurprising defaults where it's possible. This way the users don't have to provide unnecessary parameters in "happy paths" that can be guessed.
I think with this function the better way to go would be to, instead of forcing the users to input the `remote_path`parameter, it'd be better to check if it's None and then set it for the user.
This is actually already done for file uploading here https://github.com/DagsHub/client/blob/7138e44350f778a1d48ec77f36ead475d692ade6/dagshub/upload/wrapper.py#L516
But for the case of folder uploads it's not being done and it breaks.

krishnaduttPanchagnula:
@kbolashev I have added a check to see if the file path is none. Nad if it is none then we are assigning our default path 

DEFAULT_REMOTE_PATH = "/data/src/".

Please comment changes if any.

dagshub/upload/wrapper.py
```diff
@@ -194,10 +195,10 @@ def upload(
 
         """
         if os.path.isdir(local_path):
-            dir_to_upload = self.directory(remote_path)
+            dir_to_upload = self.directory(remote_path if remote_path is not None else DEFAULT_REMOTE_PATH)
             dir_to_upload.add_dir(local_path, commit_message=commit_message, **kwargs)
         else:
-            file_to_upload = DataSet.get_file(local_path, remote_path)
+            file_to_upload = DataSet.get_file(local_path, remote_path if remote_path is not None else DEFAULT_REMOTE_PATH)
             self.upload_files([file_to_upload], commit_message=commit_message, **kwargs)
```
kbolashev, line 202: The `get_file` function already handles the `remote_path is None` case. 
It gets the filename from the `local_path` and uploads with that. So this is not required + it's going to always write the file at `/data/src/`

dagshub/upload/wrapper.py
```diff
@@ -194,10 +195,10 @@ def upload(
 
         """
         if os.path.isdir(local_path):
-            dir_to_upload = self.directory(remote_path)
+            dir_to_upload = self.directory(remote_path if remote_path is not None else DEFAULT_REMOTE_PATH)
             dir_to_upload.add_dir(local_path, commit_message=commit_message, **kwargs)
```
kbolashev, line 199: This is perfectly fine and thank you for actually doing this :)

I think for this to work perfectly it'd be cool to get the directory name, and upload to `/data/{name_of_the_directory}`.

BTW, I'm going to work a little bit on the Repo class in general, so I might implement the behavior that the upload happens relative to the repo root (if the script is called in a git repository).
So, for example, if you have your repo in `~/repos/cool-repo/`, do a script with cwd set somewhere inside of the repo AND then you upload a folder `~/repos/cool-repo/some/other/folder/with/data`, we'll upload the folder to `some/other/folder/with/data`.
BUT, if cwd is not inside of a repository, then upload it to the `/data/src/` folder just like you implemented. Let me know what you think about it, I'd love to hear your thoughts :)

kbolashev (changes requested):
Leave the `is not None` check only on the directory, and then I'll merge :)
Thank you for coming back to this!!

kbolashev:
Thank you very much for the PR! 🥳

dagshub/upload/wrapper.py
krishnaduttPanchagnula: Actually thats a cool idea! @kbolashev

PR #287, opened 2023-05-15
Added Notebook Saving

jinensetpal:
Partially working for regular jupyter notebooks - only saves the session history. Works fully for Colab.


nirbarazida:
If the repo doesn't exist, it gives the following error message about the branch being missing, which doesn't guide the user in the right direction. 
```python
RuntimeError: Failed to get default branch for repository. Please specify a branch and make sure repository details are correct.
```
We should catch it, and do one of the following:
1) Present an informative error message that the repo doesn't exist
2) Ask the user if they want to open a new repo (based on the name they gave)
3) Open the repo for them

I think the first option is the best solution.

---

dagshub/nb.py
```diff
@@ -0,0 +1,44 @@
...
+
+def save(filename, path, repo_owner, repo_name, branch, commit_message='added a notebook', versioning='git') -> None:
+    """
+    IPython wrapper for saving notebooks.
+
+    :param filename (str): The name of the file under which the notebook should be saved
+    :param path (str): The path within the remote where the notebook should be saved
+    :param repo_owner (str): The username of the user who owns this repository
+    :param repo_name (str): Identify the repository
+    :param branch (str): The branch under which the notebook should be saved
```
nirbarazida, line 26: Should be by default main

PR #319, opened 2023-07-09
Fix default path on folder upload

guysmoilov:
It used to always be /data/src which is unexpected. Instead, set it to something more predictable - the relative folder path or at least its basename

### DIFF
--- a/dagshub/upload/wrapper.py
+++ b/dagshub/upload/wrapper.py
@@ -26,7 +26,6 @@ REPO_CREATE_URL = "api/v1/user/repos"
 ORG_REPO_CREATE_URL = "api/v1/org/{orgname}/repos"
 USER_INFO_URL = "api/v1/user"
 DEFAULT_DATA_DIR_NAME = 'data'
-DEFAULT_REMOTE_PATH = "/data/src/"
 logger = logging.getLogger(__name__)
 
 s = httpx.Client()
@@ -198,7 +197,15 @@ class Repo:
 
         """
         if os.path.isdir(local_path):
-            dir_to_upload = self.directory(remote_path if remote_path is not None else DEFAULT_REMOTE_PATH)
+            if remote_path is None:
+                abspath = os.path.abspath(local_path)
+                cwd = os.getcwd()
+                if abspath.startswith(cwd):
+                    remote_path = os.path.relpath(abspath, cwd)
+                else:
+                    remote_path = os.path.basename(abspath)
+            remote_path = remote_path.replace(os.sep, '/')
+            dir_to_upload = self.directory(remote_path)
             dir_to_upload.add_dir(local_path, commit_message=commit_message, **kwargs)
         else:
             file_to_upload = DataSet.get_file(local_path, remote_path)
--- a/dagshub/upload/wrapper.py
+++ b/dagshub/upload/wrapper.py
@@ -6,6 +6,7 @@ import posixpath
 import urllib
 from http import HTTPStatus
 from io import IOBase
+from pathlib import Path
 from typing import Union, Tuple, BinaryIO, Dict
 
 import httpx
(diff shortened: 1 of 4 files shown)


dagshub/upload/wrapper.py
```diff
@@ -198,7 +197,15 @@ def upload(
...
-            dir_to_upload = self.directory(remote_path if remote_path is not None else DEFAULT_REMOTE_PATH)
+            if remote_path is None:
+                abspath = os.path.abspath(local_path)
+                cwd = os.getcwd()
+                if abspath.startswith(cwd):
+                    remote_path = os.path.relpath(abspath, cwd)
+                else:
+                    remote_path = os.path.basename(abspath)
+            remote_path = remote_path.replace(os.sep, '/')
+            dir_to_upload = self.directory(remote_path)
```
kbolashev, line 208: Check out `pathlib` sometime, because some of the stuff that it can do with paths is so much nicer.
https://docs.python.org/3.7/library/pathlib.html
(Mind the 3.7, they get a lot of neat features later, but 3.7 is our baseline)

I think this code can do the same thing as what you did here, but instead of string mangling and using os libs it's all object manipulations
```python
lp = Path(local_path).resolve()
cwd = Path(".").resolve()
try:
    relative_path = lp.relative_to(cwd)
except ValueError:
    relative_path = Path(lp.name)
remote_path = relative_path.as_posix()
dir_to_upload = self.directory(remote_path)
```

tests/dda/upload/test_wrapper.py
```diff
@@ -1,26 +1,101 @@
...
+    with open(filepath2, "w") as f:
+        f.write("test data")
+
+    yield folder_path
+    try:
+        os.remove(filepath1)
+        os.remove(filepath2)
+        os.removedirs(folder_path)
+    except (OSError, FileNotFoundError):
+        pass
```
kbolashev, line 41: You can use the `tmpdir` fixture and let go of all this cleanup code, pytest will manage it for you (note the paragraph in the end, it does persist a couple test runs)
https://docs.pytest.org/en/6.2.x/tmpdir.html

tests/dda/upload/test_wrapper.py
```diff
@@ -1,26 +1,101 @@
...
     yield filepath
     try:
         os.remove(filepath)
-    except OSError:
+    except (OSError, FileNotFoundError):
         pass
 
 
-def test_upload_dataset_closes_files(mock_api, upload_repo: Repo, test_file: str):
+@pytest.fixture(scope="function")
```
kbolashev, line 23: Default scope is function
```suggestion
@pytest.fixture
```

tests/dda/upload/test_wrapper.py
```diff
@@ -1,26 +1,101 @@
...
+    os.makedirs(folder_path, exist_ok=True)
+
+    filepath1 = os.path.join(folder_path, f"{uuid.uuid4()}.txt")
+    with open(filepath1, "w") as f:
+        f.write("test data")
+    filepath2 = os.path.join(folder_path, f"{uuid.uuid4()}.txt")
+    with open(filepath2, "w") as f:
+        f.write("test data")
+
+    yield folder_path
```
kbolashev, line 35: You can return a Path object of it and make your life easier later re: file creation

kbolashev (approved):
LGTM, comments are mostly just neat possible timesavers

tests/dda/upload/test_wrapper.py
guysmoilov: That's nice but if it's alright with you I'd rather avoid it.
It's waaaay too magical. Basically impossible to figure out where it's coming from.

kbolashev: for sure, just a thing to check out

dagshub/upload/wrapper.py
guysmoilov: I tried but so far it's having edge cases.
Namely, if `local_path` is just a non-absolute path, then `local_path.relative_to(Path.cwd())` fails with `ValueError` and ruins everything.
I guess because it requires using `join` but that makes everything way more complicated

tests/dda/upload/test_wrapper.py
guysmoilov: How so?

PR #324, opened 2023-07-20
Fix for mirror upload sync issues

kbolashev:
This should mitigate the problems with doing multiple uploads to a mirror repo.
Works on a Repo object, so as long as the object is reused, the uploads should be consistent

It:
- Polls for the new last commit every second
- After 10 minutes of timeout logs a warning and tries to push anyway.
- If during upload a 204 (No Content) is received, then consider that there was nothing uploaded and be ready for the next push immediately


dagshub/upload/wrapper.py
```diff
@@ -202,6 +204,10 @@ def __init__(
 
         self._api = RepoAPI(f"{owner}/{name}", host=self.host, auth=self.auth)
 
+        # For mirror uploading: store the last revision for which we uploaded
+        # When the last revision changes, that means the sync has been complete and we can upload a new batch
+        self._last_upload_revision: Optional[str] = None
+
```
guysmoilov, line 210: Putting this as a variable on the `Repo` object feels wrong - it will cause shenanigans for any user trying to use `Repo.upload_files` directly. It should be a variable in whichever for loop is splitting the upload into chunks.

guysmoilov (changes requested):
Sorry I think this impl strategy is dangerous, any reason not to do this as a variable in the loop?

dagshub/upload/wrapper.py
kbolashev: But this way the upload works across the whole repo object.
If the user is, say, training stuff and uploading model results in a loop (new artifact after each loop step), you can't guarantee sync consistency there

kbolashev: This is doing exactly the opposite of causing shennanigans.
If I as a user that doesn't upload the whole folder at once, do:
```
repo = Repo("user", "mirror-repo")
repo.upload("file1.txt")
repo.upload("file2.txt")
```
We will break

guysmoilov: Hmm I guess since the branch is part of the `Repo` object and I can't upload to different branches in parallel, you're probably right...
This makes me think of a different issue - if new_branch is passed, then no check needs to be made

kbolashev: Good catch, thank you! I'll fix it

kbolashev: I guess it might be a good idea to put a warning that this class is overall not thread safe.

dagshub/upload/wrapper.py
```diff
@@ -281,6 +290,11 @@ def upload_files(
                     "new_branch_name": new_branch,
                 }
             )
+            # Drop the upload revision because we're sure this is a new branch
+            self._last_upload_revision = None
```
guysmoilov, line 294: Actually this isn't how I would do it.
Let's look at this scenario:

```python
repo = Repo('a', 'b', branch='c')
repo.upload_files(...)
repo.upload_files(..., new_branch='d')
repo.upload_files(...)
```

The 3rd upload_files can still benefit from the `_last_upload_revision`

Instead, and I think that would also be more readable - if the `new_branch` is passed, then totally skip all the checks and calls to `self.poll_mirror` etc.

Then the repo object is still good for more uploads, and it should be safe as you're working on a totally new branch

dagshub/upload/wrapper.py
```diff
@@ -294,6 +306,9 @@ def upload_files(
             timeout=None,
         )
         self._log_upload_details(data, res, files)
+        # 204 means nothing was added, so if we're in the mirror we can upload the next batch immediately
+        if self._api.is_mirror and res.status_code == 204:
```
guysmoilov, line 310:
```suggestion
        if new_branch is None and self._api.is_mirror and res.status_code == 204:
```

guysmoilov (changes requested):

dagshub/upload/wrapper.py
kbolashev: Right, sorry, you're right

guysmoilov (approved):

PR #327, opened 2023-07-23
Fix URL checking of create_datasource

guysmoilov:
Source of the problem was that `None != ""`.

I'd be happy to learn how to effectively write a unit test for this next time.

### DIFF
--- a/dagshub/data_engine/datasources.py
+++ b/dagshub/data_engine/datasources.py
@@ -106,6 +106,3 @@ __all__ = [
     get,
     get_or_create,
 ]
-
-if __name__ == '__main__':
-    create_datasource('sasi', 'adg', 'adgadg')
--- a/dagshub/data_engine/datasources.py
+++ b/dagshub/data_engine/datasources.py
@@ -1,13 +1,11 @@
 import logging
-import urllib.parse
 from typing import Optional, Union, List
 
 from dagshub.common.analytics import send_analytics_event
 from dagshub.common.api.repo import RepoAPI
 from dagshub.data_engine.client.data_client import DataClient
-from dagshub.data_engine.client.models import DatasourceType
 from dagshub.data_engine.model.datasource import Datasource
-from dagshub.data_engine.model.datasource_state import DatasourceState
+from dagshub.data_engine.model.datasource_state import DatasourceState, DatasourceType, path_regexes
 from dagshub.data_engine.model.errors import DatasourceNotFoundError
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


dagshub/data_engine/datasources.py
```diff
@@ -30,18 +30,19 @@ def create_datasource(repo: str, name: str, path: str, revision: Optional[str] =
...
-        and parsed.query == "" \
-        and parsed.params == "" \
-        and parsed.fragment == "":
+    if parsed.path \
+       and not parsed.scheme \
+       and not parsed.hostname \
+       and not parsed.query \
+       and not parsed.params \
+       and not parsed.fragment:
         return create_from_repo(repo, name, path=parsed.path, revision=revision)
```
kbolashev, line 39: I'm not sure that this will _actually_ work, because what happens when I want to create a datasource from a repo path "www.google.com/hello"?
I'm a wacky little guy and decided to name my folders like that (I'm scraping websites maybe).
This will make us think this is a bucket URL, when in actuality it is a repo path.

I think that the way to go here should be to check against a regex that I already wrote. Sorry for not mentioning it on the last review.

https://github.com/DagsHub/client/blob/a6a757a1d0805b4551552989928498cf2355ee28/dagshub/data_engine/model/datasource_state.py#L22

dagshub/data_engine/datasources.py
```diff
@@ -30,18 +30,19 @@ def create_datasource(repo: str, name: str, path: str, revision: Optional[str] =
...
 
-    elif parsed.scheme != "" and parsed.hostname != "":
+    elif parsed.scheme and parsed.hostname:
         # Bucket URL
-        if parsed.query != "" or parsed.params != "" or parsed.fragment != "":
+        if parsed.query or parsed.params or parsed.fragment:
             raise ValueError("Invalid bucket URL: ", path)
-        if revision != "":
+        if revision:
```
kbolashev, line 45: I checked what `urllib.parse` is returning, and it's returning empty strings on all fields. No Nones anywhere.
So that means that this one line is the actual line that broke stuff. I'd just check that the revision is not set, everything else is basically user error (although maybe it's a good idea for _us_ to check that a branch even exist)

```suggestion
        if revision is not None:
```

kbolashev (changes requested):

dagshub/data_engine/datasources.py
guysmoilov: For me it returned `None`s on some fields and it caused the whole function to.... malfunction ¯\_(ツ)_/¯

dagshub/data_engine/datasources.py
guysmoilov: So you're saying that if it doesn't match the bucket regex, then assume it's a path?
I think your regex also won't work, e.g. if there are any dots in the bucket name?

dagshub/data_engine/datasources.py
guysmoilov: The hostname, at least, is also `None`. The whole thing seems scarily inconsistent.
image

guysmoilov: Also, I don't want to check `revision is not None` - if it's an empty string, I feel comfortable just throwing it out

dagshub/data_engine/datasources.py
kbolashev: I checked online while I wrote this regex and while it's ok to have dots in the bucket name, it apparently breaks a lot of stuff so it's not recommended.
Thinking about it, I probably shouldn't be that opinionated about users' bucket names, so maybe it's worth changeing.

But yea, the idea is check against the regex, if it passes then it's a bucket (need to be a full match though, I see that I don't have `^` and `$` in these, which might be a mistake?)

I did write some tests for this, so you can change the regex to whatever you want as long as they pass: https://github.com/DagsHub/client/blob/a6a757a1d0805b4551552989928498cf2355ee28/tests/data_engine/test_datasource_state.py

guysmoilov: Or underscores

dagshub/data_engine/datasources.py
kbolashev: > The hostname, at least, is also `None`. The whole thing seems scarily inconsistent.
Yea, that whole part of the core library is very weird IMO. Especially after how well paths work now, urls feel really bad now

dagshub/data_engine/datasources.py
```diff
@@ -120,3 +106,6 @@ def _create_datasource_state(repo: str, name: str, source_type: DatasourceType,
     get,
     get_or_create,
 ]
+
+if __name__ == '__main__':
+    create_datasource('sasi', 'adg', 'adgadg')
```
guysmoilov, line 111:
```suggestion
```

kbolashev (approved):
Negative LOC change 🥳 love to see that

PR #332, opened 2023-07-26
External bucket support for file_column detection, dataloader post_hooks, bugfixes

jinensetpal:
Bugfix: Handled RepoApi single-file edge case for external buckets


dagshub/data_engine/client/loaders/base.py
```diff
@@ -81,7 +81,22 @@ def _get_file_columns(self):
             [self.entries[0].metadata[col] for col in self.metadata_columns],
         ):
             try:
-                self.repo.list_path((self.datasource_root / str(value)).as_posix())
+                if self.source == "repo":
+                    self.repo.list_path((self.datasource_root / str(value)).as_posix())
```
kbolashev, line 85: Replace the `self.source` in this class to use the `self.datasource.source.source_type` enum instead of splitting the path. You can compare against the enum then

dagshub/data_engine/client/loaders/base.py
```diff
@@ -81,7 +81,22 @@ def _get_file_columns(self):
...
+                            Path(
+                                "/".join(
+                                    list(self.datasource.source.path_parts().values())[
+                                        :2
+                                    ]
+                                )
+                            )
+                            / self.datasource_root
+                            / str(value)
+                        ).as_posix()
```
kbolashev, line 98: I think I already did this same thing once.
Check out `self.datasource.source.root_content_path`. This will give you `https://dagshub.com/api/v1/repos/jinen/test-repo/storage/content/s3/bucket/`, and then you'll need to add the path of the datapoint itself to it.
The function also handles repos and buckets, so no need for the if/else
https://github.com/DagsHub/client/blob/11f90c340779b7d4d4fb584b8c86662cbb10f556/dagshub/data_engine/model/datasource_state.py#L124

dagshub/common/api/repo.py
```diff
@@ -175,7 +175,10 @@ def _get():
                 logger.debug(res.content)
                 raise RuntimeError(error_msg)
 
-            return dacite.from_dict(StorageContentAPIResult, res.json())
+            content = res.json()
+            if 'entries' not in content:
+                content = {'entries': [content, ], 'next_token': None}
+            return dacite.from_dict(StorageContentAPIResult, content)
```
kbolashev, line 181: Good catch. I always forget about this behavior, so thank you for noticing it

kbolashev (changes requested):

dagshub/data_engine/client/loaders/base.py
kbolashev: Wait, even better, there's a `self.datasource.source.content_path`, that also appends the path!
https://github.com/DagsHub/client/blob/11f90c340779b7d4d4fb584b8c86662cbb10f556/dagshub/data_engine/model/datasource_state.py#L89

jinensetpal: Awesome, thank you!

jinensetpal: I just realized, this won't actually work, since the idea behind it is not getting the value, but checking if it exists in the remote. Those functions get me the target url, but obtaining the final path (just from 's3/bucket/path/to/file') still requires string manipulation similar to what's already on.

I cannot directly pass the URL to `list_storage_path` or `list_path` in RepoAPI, and I do not want to send queries within the dataloader class because of DRY and I'll have to manage authentication.

kbolashev: The function I wrote just generates the path though, it doesn't do a request. Can you demonstrate an example where you know it won't work for some reason?

dagshub/data_engine/client/loaders/torch.py
```diff
@@ -18,13 +18,42 @@ def __init__(self, *args, **kwargs):
...
+    def __init__(self, *args, post_hook, **kwargs):
+        self.post_hook = post_hook
+        super().__init__(*args, **kwargs)
+
+
 class PyTorchDataLoader(torch.utils.data.DataLoader):
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args, post_hook=lambda x: x, **kwargs):
         super().__init__(*args, **kwargs)
+        self.post_hook = post_hook
```
kbolashev, line 45: What's the purpose of the post hook?
I don't see any code in here that has custom post-hooks, so I assume this is accomodating for something you wrote yourself.
Can you add some docs?

dagshub/data_engine/client/loaders/tf.py
```diff
@@ -22,14 +22,22 @@ def __init__(self, *args, **kwargs):
 
     def generator(self):
         for idx in range(len(self)):
-            yield self[idx]
+            yield tuple(self[idx])
```
kbolashev, line 25: I really don't understand much of how TF/PT works and maybe it's time for me to investigate, but why are you changing it to return tuples now?

kbolashev (changes requested):
I'm getting lost in this code now if I can be honest.
Do you want to maybe try writing some unit tests together for this? Because I doubt anyone coming into this code fresh that will have to maintain it in the future would have any idea what's going on.

kbolashev:
Merging it in skipping the comments because we need it sooner rather than later

PR #348, opened 2023-08-27
Add enrichment tag annotation

idonov8:


dagshub/data_engine/dtypes.py
```diff
@@ -0,0 +1,37 @@
...
+        return self
+
+    def is_annotation(self):
+        return ReservedTags.ANNOTATION in self.tags
+
+    def tag(self, tag):
+        if tag not in ReservedTags._value2member_map_:
+            self.tags.append(tag)
+            return self
+        raise ValueError("This tag is reserved, choose a different tag name")
```
kbolashev, line 22: Very bad idea to rely on private members like that. 
Can't guarantee it won't disappear at some point, so only do this type of stuff as a last resort when you are out of other options

You can do the same thing here by doing a test assignment:
```python
try:
  val = ReservedTags(tag)
  is_reserved = True
except ValueError:
  is_reserved = False
```

dagshub/data_engine/model/query.py
```diff
@@ -14,10 +15,15 @@
...
+    Int: MetadataFieldType.INTEGER,
     bool: MetadataFieldType.BOOLEAN,
+    Bool: MetadataFieldType.BOOLEAN,
     float: MetadataFieldType.FLOAT,
+    Float: MetadataFieldType.FLOAT,
     str: MetadataFieldType.STRING,
+    String: MetadataFieldType.STRING,
     bytes: MetadataFieldType.BLOB,
+    Blob: MetadataFieldType.BLOB,
 }
```
kbolashev, line 27: I think this will break the reverse lookup table, because it will only have one of the types now (since the reverse is also a dict).

I think the "clean" way to solve this is to add a `classmethod` onto the dtypes that will return the MetadataFieldType, and when you want to check for the type do a `if isinstance(value, DagshubDataType)`

Hope it makes sense

kbolashev (commented):
sorry for coming in like that, I see that it's still a draft, but still there are a couple things that need a fix

PR #351, opened 2023-09-03
feature/support azure datasource

simonlsk:
# Spec
- [x] Allow datasource to be stored on azure blob
- [x] Allow streaming from azure
- [x] Allow dedicated azure downloader

### DIFF
--- a/dagshub/common/download.py
+++ b/dagshub/common/download.py
@@ -73,6 +73,31 @@ def enable_s3_bucket_downloader(client=None):
     add_bucket_downloader("s3", get_fn)
 
 
+def enable_azure_container_downloader(account_url, client=None):
+    """
+    Enables downloading storage items using the azure client, instead of going through DagsHub's server.
+    For custom clients use `enable_custom_bucket_downloader` function.
+
+    Args:
+        account_url: an azure storage account url, of the form "https://<storage-account-name>.blob.core.windows.net"
+        client: a boto3 S3 client.
+            If client isn't specified, the default parameterless constructor is used.
+            If specified, account_url is disregarded, and the client is used.
+    """
+    import io
+    if client is None:
+        from azure.storage.blob import BlobServiceClient
+        from azure.identity import DefaultAzureCredential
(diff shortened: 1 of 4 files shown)


simonlsk:
[Add Azure to Python client](https://trello.com/c/TuAHyvc2/1884-add-azure-to-python-client)

dagshub/common/download.py
```diff
@@ -73,6 +73,32 @@ def get_fn(bucket, path) -> bytes:
...
+        client: a boto3 S3 client.
+            If client isn't specified, the default parameterless constructor is used.
+            If specified, account_url is disregarded, and the client is used.
+    """
+    import io
+    if client is None:
+        from azure.storage.blob import BlobServiceClient
+        from azure.identity import DefaultAzureCredential
+        client = BlobServiceClient(account_url, credential=DefaultAzureCredential())
+
```
kbolashev, line 92: Problems with this snippet: If client WAS specified, you still need to enter your account_url for some reason.

Way to solve it: make `account_url` optional, have an assert at the start that either of account_url or client is not None

The DefaultAzureCredentials looks weird to me, but I don't see a reason for exposing yet another arg, so it's fine with me

dagshub/common/download.py
```diff
@@ -73,6 +73,32 @@ def get_fn(bucket, path) -> bytes:
...
 
 
+def enable_azure_container_downloader(account_url, client=None):
+    """
+    Enables downloading storage items using the azure client, instead of going through DagsHub's server.
+    For custom clients use `enable_custom_bucket_downloader` function.
+
+    Args:
+        account_url: an azure storage account url, of the form "https://<storage-account-name>.blob.core.windows.net"
+        client: a boto3 S3 client.
```
kbolashev, line 83:
```suggestion
        client: preconfigured azure.storage.blob.BlobServiceClient
```

kbolashev (changes requested):

kbolashev (approved):

PR #354, opened 2023-09-07
Bugfix: Make saving notebooks work on Windows

kbolashev:
Also added a warning about the way saving notebooks in Jupyter works


kbolashev:
[Fixing save_notebook](https://trello.com/c/o2hN3WpR/1895-fixing-savenotebook)

PR #357, opened 2023-09-18
generate and save permanent token after going through oauth

simonlsk:
Provided OAuth token sometimes ends in the middle of training, then starts throwing 401’s. This happens also for users that use local notebooks, where they work intermittently.

### Proposed solution
After oauth token is triggered, set a permanent token for the user using the token API, unless specified otherwise.

### DIFF
--- a/dagshub/auth/tokens.py
+++ b/dagshub/auth/tokens.py
@@ -3,8 +3,11 @@ import logging
 import os
 import threading
 import traceback
+import urllib
+from http import HTTPStatus
 from typing import Optional, Dict, List, Set, Union
 
+import requests
 import yaml
 from httpx import Auth
 
@@ -18,6 +21,7 @@ from dagshub.common.util import multi_urljoin
 logger = logging.getLogger(__name__)
 
 APP_TOKEN_TYPE = "app-token"
+CLIENT_TOKEN_NAME = "client"
 
 
 class InvalidTokenError(Exception):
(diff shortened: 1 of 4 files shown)


kbolashev (approved):
:(

dagshub/auth/tokens.py
```diff
@@ -360,3 +371,24 @@ def add_oauth_token(host: Optional[str] = None, **kwargs):
...
+    # todo: if a permanent token already exists and is valid, skip the next step
+    perm = _generate_permanent_token(token)
+    _get_token_storage(**kwargs).add_token(perm, host, skip_validation=True)
+
+
+def _generate_permanent_token(auth: DagshubTokenABC) -> str:
+    host = config.host
+    tokens_path = "api/v1/user/tokens"
+    tokens_url = urllib.parse.urljoin(host, tokens_path)
+    token_data = {"name": CLIENT_TOKEN_NAME}
```
kbolashev, line 387: Add a date in there or something like that

dagshub/auth/tokens.py
```diff
@@ -360,3 +371,24 @@ def add_oauth_token(host: Optional[str] = None, **kwargs):
     host = host or config.host
     token = oauth.oauth_flow(host)
     _get_token_storage(**kwargs).add_token(token, host, skip_validation=True)
+
+
+def add_permanent_token(host: Optional[str] = None, **kwargs):
+    token = get_token_object(host=host, **kwargs)
+    # todo: if a permanent token already exists and is valid, skip the next step
+    perm = _generate_permanent_token(token)
```
kbolashev, line 379: Check that `token` is of type `OAuthDagshubToken`, that should be enough
Alternatively add an `is_expirable()` onto the Token ABC, but that's probably overkill

dagshub/auth/tokens.py
```diff
@@ -163,11 +170,15 @@ def get_token_object(self, host: str = None, fail_if_no_token: bool = False, **k
...
+                        permanent_token = _generate_permanent_token(good_token)
+                        if permanent_token is None:
+                            logger.warning("Could not generate permanent token")
+                        tokens.append(permanent_token)
                     # Save the cache
                     self._token_cache[host] = tokens
                     self._store_cache_file()
-
+                    # Use oauth token to set a permanent token if not disabled
             return good_token
```
kbolashev, line 182: Forgot to implement?

kbolashev (changes requested):

simonlsk:
Oauth slider in the redemption screen will let users select how long the token should last. PR no longer necessary.

PR #360, opened 2023-09-19
Fix problems with blob field filter deserialization

kbolashev:
Before this PR deserializing a query with a blob field filter on it would throw the following error:
```
TypeError: string argument without an encoding
```
Because it was trying to call `bytes("b'<whatever_value_was_there>'")`

This PR fixes this behavior and correctly deserializes the binary values. As a fallback if I don't get the `b'`/`b"`-prefix, I turn the whole value into the byte array

Tests for the custom deserializer + a loop of deserialize->serialize included

### DIFF
--- a/dagshub/data_engine/model/query.py
+++ b/dagshub/data_engine/model/query.py
@@ -20,8 +20,17 @@ _metadataTypeLookup = {
     bytes: MetadataFieldType.BLOB,
 }
 
+
+def bytes_deserializer(val: str) -> bytes:
+    if val.startswith('b"') or val.startswith("b'"):
+        return val[2:-1].encode()
+    # Fallback - encode whatever we got from the server
+    return val.encode()
+
+
 _metadataTypeCustomConverters = {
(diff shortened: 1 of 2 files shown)


simonlsk (changes requested):
Thanks, this will solve the current issue with deserialization!
I think we should also have a dedicated handling for bytes at serialization time. 
WDYT?

kbolashev:
> Thanks, this will solve the current issue with deserialization! I think we should also have a dedicated handling for bytes at serialization time. WDYT?

Sure, what do we want to do with those? gzip -> base64, same as uploading?
Means I'll have to also do the reverse here then

simonlsk:
> Sure, what do we want to do with those? gzip -> base64, same as uploading?
Means I'll have to also do the reverse here then

Sorry, I mean when running the `saveDataset` mutation, the value in `filter.query.filter.value` should be an empty string when the field type is blob, and the query is `is_null`.

simonlsk (approved):
LGTM, thank you!

PR #362, opened 2023-09-26
DDA Streaming: allow excluding certain glob patterns from being looked up

kbolashev:
That way, the users can control what to do with the files that don't exist on the repo, but they don't want their framework to still try to look them up.

Yono stumbled upon this problem while training a YOLO model. It kept looking up `*.npy` files trying to find if they were cached, which led to a lot of unnecessary requests to the server and general slowness of the experience.

Also fixed docstrings in wrapper.py, PyCharm wasn't picking them up correctly


kbolashev:
[Ability to exclude glob patterns from being looked up on server in DDA Streaming](https://trello.com/c/gd8tZ8X5/1907-ability-to-exclude-glob-patterns-from-being-looked-up-on-server-in-dda-streaming)

Issue #353, opened 2023-09-05
Query compositing results in unexpected bug


yonomitt:
If I subquery an already queried `Datasource`, the subquery is not properly composited with the original query.

For instance:

<img width="563" alt="Screenshot 2023-09-05 at 10 47 35" src="https://github.com/DagsHub/client/assets/8521453/fdafeef2-8f29-476d-bb68-ecfb7f1614c3">

The first query filters the `Datasource` down to 539 items, but the 2nd level queries ignore this (there are a total of 1008 items in the unfiltered `Datasource`)

The actually query description look like this:

<img width="702" alt="Screenshot_2023-09-05_at_11 15 06" src="https://github.com/DagsHub/client/assets/8521453/00d84140-9746-4261-a1da-2b59599a1bb9">

# Workaround:

If you use this notation, it works:

```
labeled = ds[ds['has_annotation'] == True]

train = labeled[labeled['split'] == 'train']
```

kbolashev:
![](https://github.trello.services/images/mini-trello-icon.png) [Fix sequential querying breaking in the client](https://trello.com/c/WGcJ4dhW/1924-fix-sequential-querying-breaking-in-the-client)

kbolashev:
Should be fixed now :)

PR #374, opened 2023-10-05
Fix sequential querying composition bug

kbolashev:
Solves issue #353 

Added test case for the issue.
This introduces a breaking change, because now the ds[ds[..]] composition works differently: Now instead of composing the outer datasource with the inner one, the client uses the query of the inner datasource.
I imagine this is a more expected behavior for a lot of people, than composing them


kbolashev:
[Fix sequential querying breaking in the client](https://trello.com/c/WGcJ4dhW/1924-fix-sequential-querying-breaking-in-the-client)

PR #381, opened 2023-11-02
Move jupyter + fuse into optional dependencies

kbolashev:
Moved the jupyter fixes for rich + fuse library into their own optional dependencies.
Now installing the jupyter dep is `pip install dagshub[jupyter]`

Tested this branch in Colab, everything kind of works still, but rich's progress bars don't show up until the last second. Doesn't break functionality, but still will probably need to change the notebooks to `pip install dagshub[jupyter]`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ install_requires = [
     "click>=8.0.4",
     "httpx~=0.23.0",
     "GitPython>=3.1.29",
-    "rich[jupyter]~=13.1.0",
+    "rich~=13.1.0",
     # Need to keep dacite version in lockstep with voxel, otherwise stuff breaks on their end
     "dacite~=1.6.0",
     "tenacity~=8.2.2",
@@ -42,6 +42,11 @@ install_requires = [
     "pathvalidate~=3.0.0",
     "python-dateutil",
 ]
+
+extras_require = {
+    "jupyter": ["rich[jupyter]~=13.1.0"],
+    "fuse": ["fusepy>=3"],
+}
 
 # Polyfills for Python 3.7
 if sys.version_info.major == 3 and sys.version_info.minor == 7:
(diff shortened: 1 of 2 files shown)


kbolashev:
I tried to look into moving the whole project structure to `pyproject.toml`, but it seemed like we won't get any advantages from it, aside from being on "the cool new stuff", so I decided to not do that for now

setup.py
```diff
@@ -31,7 +31,7 @@ def get_version(rel_path: str) -> str:
     "click>=8.0.4",
     "httpx~=0.23.0",
     "GitPython>=3.1.29",
-    "rich[jupyter]~=13.1.0",
+    "rich~=13.1.0",
```
guysmoilov, line 34: So the regular rich package just works without modification?
Cool

guysmoilov (approved):
Should be OK overall, but:
1. If the experience in Jupyter is now completely broken (no progress bar appearing until it's finished), is there any reasonable workaround we could do? Abstraction around rich, and use something different if we're in a notebook and are missing the dependencies? How hard would that be?
2. Need to modify the readme

setup.py
kbolashev: The regular rich still works for CLI. In Jupyter it's broken though until you install the [jupyter] optional dep.

We could check in the `install.py` script if it's being executed inside a jupyter env/if there is jupyter installed. But from what I've read this is not recommended, because e.g. if we were to ever build wheels, then the wheel doesn't execute the setup.py, and I think the result would be that `rich[jupyter]` will always be a dependency. At least that's what I read in some places. It is all surprisingly underdocumented and underexplained how it works

kbolashev: BUT right now we could just put an `is_in_notebook()` check inside of setup.py that tries to `import IPython` and change the installed rich based on that

PR #383, opened 2023-11-06
Bug: setting annotation metadata two times doesn't work

kbolashev:
Reproduction:

```python
ds.metadata_field("annotation").set_annotation().apply()   # Works
ds.metadata_field("annotation").set_annotation().apply()   # Gets GQL error because we're sending double annotation tag
```

Replaced the metadata field tag list with a set


idonov8 (approved):
Looks great! Thank you

PR #386, opened 2023-11-12
versioning

sdafni:
TODO @kbolashev suggestions: 

1. currently there is not verifying of the select list. i.e it can include

- duplication: (FIeld('x'), Field('x'))
- contradiction: (FIeld('x', as_of_time=t1, alias='x_t1'), FIeld('x', as_of_time=t2, alias='x_t1')
- missed aliases (FIeld('x', as_of_time=t1), FIeld('x', as_of_time=t2))

2.  if select +include_all is used, we'll get all columns. its implemented in client by adding all names. would be more efficient to do it in backend

3. Guy suggests '*' as give me all, otherwise only selected


dagshub/data_engine/model/datasource.py
```diff
@@ -120,6 +130,20 @@ def all(self) -> "QueryResult":
         self._check_preprocess()
         return self._source.client.get_datapoints(self)
 
+    def select(self, selected: List[Field]):
+        self._select = []
```
kbolashev, line 134: Add a docstring explaining the purpose of the function + a usage example

dagshub/data_engine/model/datasource.py
```diff
@@ -61,13 +61,20 @@ class DatapointMetadataUpdateEntry(json.JSONEncoder):
     allowMultiple: bool = False
 
 
+@dataclass
+class Field:
+    column: str
+    as_of_timestamp: Optional[float] = None
+    alias: Optional[str] = None
```
kbolashev, line 68: This needs to handle both raw timestamps AND datetime objects

Right now it looks like the **only** way to have an as_of right now is to filter directly with the timestamp, which is certainly not user-friendly. We should be able to handle this:
```
as_of_date = datetime.now() - timedelta(days=1)
ds.select(Field("blabla", as_of=as_of_date))
```

How to implement:
Basically the `as_of` needs to be a `Optional[Union[float, datetime]]` and the `as_of_timestamp` will be a property that will have special sauce to turn the `datetime` into a float.

tests/data_engine/test_querying.py
```diff
@@ -27,6 +27,25 @@ def test_simple_filter(ds):
...
+                                                          {'gt': {'data': {'field': 'x', 'as_of': 345, 'value': 2}}}],
+                                             'data': None}},
+                                    {'gt': {'data': {'field': 'y', 'as_of': 789, 'value': 3}}}], 'data': None}}
+    assert q.to_dict() == expected
+
+    expected_select = [{'name': 'y', 'asOf': 123}, {'name': 'x', 'asOf': 456, 'alias': 'y_t1'}]
+    assert ds2._select == expected_select
+
+
 def test_composite_filter(ds):
```
kbolashev, line 49: After you implement the datetime handling, add tests here that ONLY handle the datetime stuff, don't overdo it with a complex composition query, just one select should be enough

tests/data_engine/test_querying.py
```diff
@@ -27,6 +27,25 @@ def test_simple_filter(ds):
     assert q.to_dict() == expected
 
 
+def test_versioning(ds):
+    add_int_fields(ds, "x")
+    add_int_fields(ds, "y")
+
+    ds2 = ((ds[ds[Field("x", as_of_timestamp=123.99)] > 1]) &
+           (ds[ds[Field("x", as_of_timestamp=345)] > 2]) |
+           (ds[ds[Field("y", as_of_timestamp=789)] > 3])).\
```
kbolashev, line 36: If you want to have this be an example of a "complex" query for versioning, I'd switch the second X for a non-timestamped just for giggles, but that's up to you

dagshub/data_engine/model/datasource.py
kbolashev: Additionally: change the signature of the function to be
```suggestion
    def select(self, *selected: Field):
        self._select = []
```
Skips you from creating a list in the arguments

dagshub/data_engine/model/datasource.py
```diff
@@ -120,6 +130,20 @@ def all(self) -> "QueryResult":
...
 
+    def select(self, selected: List[Field]):
+        self._select = []
+
+        for s in selected:
+            new_select_field = {"name": s.column}
+            if s.as_of_timestamp:
+                new_select_field["asOf"] = int(s.as_of_timestamp)
+            if s.alias:
+                new_select_field["alias"] = s.alias
```
kbolashev, line 141: Turn this into a `to_dict()` function on the Field object
Then this whole function can be turned into
```
self._select = [s.to_dict() for s in selected]
```

tests/data_engine/test_querying.py
kbolashev: Also add a test for the dataset deserialization. Validate that if the recieved dict has a select field, that the select field shows up in the datasource.
It should be self evident, but I want it to be there just to be sure later. The serialization/deserialization stuff is changing a lot, so I want to be sure it has good test coverage

kbolashev (changes requested):

dagshub/data_engine/model/query.py
```diff
@@ -153,14 +161,17 @@ def _serialize_node(node: Node, tree: Tree) -> Dict:
...
                     f"Supported types: {list(metadataTypeLookup.keys())}"
                 )
-            return {
+            res = {
                 "filter": {
                     "key": key,
                     "value": str(value),
                     "valueType": value_type,
-                    "comparator": query_op.value,
+                    "comparator": query_op.value
```
kbolashev, line 169: Leave the comma here please

dagshub/data_engine/model/query.py
```diff
@@ -192,7 +203,8 @@ def _deserialize_node(node_dict: Dict, tree: Tree, parent_node=None) -> None:
             value_type = metadataTypeLookupReverse[val["valueType"]]
             converter = _metadataTypeCustomConverters.get(value_type, lambda x: value_type(x))
             value = converter(val["value"])
-            node = Node(tag=comparator, data={"field": key, "value": value})
+            as_of = val.get("asOf")
+            node = Node(tag=comparator, data={"field": key, "value": value, "as_of": as_of})
```
kbolashev, line 207: Are we handling `as_of = None` correctly? Because it looks like it will always be applied in this code

sdafni: there is a final check in _serialize_node

tests/data_engine/test_querying.py
sdafni: 1. not sure i understand, i added a test to the datetime/timestamp options its inside my existing test
2. done

kbolashev: I want to have extremely basic true unit-tests that test one thing and one thing only.
A more complex test that tests interactions is also extremely valid, but if something changes in the code for handling e.g. specifically datetimes, the whole complex test fails, but it will still require combing to understand what exactly failed.
If we have a very barebones and simple test that specifically handles a timestamp in one case, and a datetime in another, any failing change would be quicker to triage

dagshub/data_engine/model/datasource.py
```diff
@@ -61,13 +62,38 @@ class DatapointMetadataUpdateEntry(json.JSONEncoder):
...
+                return int(self.as_of_time.timestamp())
+            else:
+                return int(self.as_of_time)
+        else:
+            return None
+
+    def to_dict(self) -> Dict[str, Any]:
+        res_dict = {"name": self.column}
+        if self.as_of_time:
+            res_dict["asOf"] = self.as_of_timestamp
```
kbolashev, line 84: Filtering from the 0 timestamp should still be valid, even if it doesn't make sense
```suggestion
        if self.as_of_time is not None:
            res_dict["asOf"] = self.as_of_timestamp
```

tests/data_engine/test_querying.py
```diff
@@ -27,6 +28,42 @@ def test_simple_filter(ds):
...
 
+def test_versioning(ds):
+    add_int_fields(ds, "x")
+    add_int_fields(ds, "y")
+    add_int_fields(ds, "z")
+
+    ds2 = ((ds[ds[Field("x", as_of_time=123.99)] > 1]) &
+           (ds[ds[Field("x", as_of_time=345)] > 2]) |
+           (ds[ds[Field("y", as_of_time=789)] > 3])).\
+        select(Field("y", as_of_time=123), Field("x", as_of_time=456, alias="y_t1"), Field("z", as_of_time=dateutil.parser.parse("Wed 22 Nov")))
```
kbolashev, line 39: This will need to be changed every year.
Hardcode one specific value with a boring `datetime.datetime()` constructor.
You can even do a [datetime.fromtimestamp](https://docs.python.org/3/library/datetime.html#datetime.datetime.fromtimestamp) and reuse the same value as the query in the result (set the tz to `timezone.utc` too so there's no localtime issues)

sdafni: addes year

tests/data_engine/test_querying.py
```diff
@@ -27,6 +28,57 @@ def test_simple_filter(ds):
...
+    # timestamp
+    ds2 = ds[ds[Field("x", as_of_time=1700604000)] > 1]
+    q = ds2.get_query()
+    assert q.to_dict() == {'gt': {'data': {'as_of': 1700604000, 'field': 'x', 'value': 1}}}
+
+    # datetime
+    ds2 = ds[ds[Field("x", as_of_time=dateutil.parser.parse("Wed 22 Nov 2023"))] > 1]
+    q = ds2.get_query()
+    assert q.to_dict() == {'gt': {'data': {'as_of': int(dateutil.parser.parse("Wed 22 Nov 2023").timestamp()),
+                                           'field': 'x', 'value': 1}}}
```
kbolashev, line 43: Separate these into two different tests.
This way if datetime serialization fails, you see X on the datetime test, but not on the timestamp one

dagshub/data_engine/model/datasource.py
```diff
@@ -61,13 +62,38 @@ class DatapointMetadataUpdateEntry(json.JSONEncoder):
...
 
+@dataclass
+class Field:
+    column: str
+    as_of_time: Optional[Union[float, datetime.datetime]] = None
+    alias: Optional[str] = None
+
+    @property
+    def as_of_timestamp(self) -> Optional[int]:
+        if self.as_of_time:
```
kbolashev, line 73: Same suggestion as it was in the `to_dict`:
```suggestion
        if self.as_of_time is not None:
```

dagshub/data_engine/datasets.py
```diff
@@ -41,7 +40,7 @@ def _from_gql_result(repo: str, dataset_result: "DatasetResult") -> "Datasource"
     ds = Datasource(DatasourceState.from_gql_result(repo, dataset_result.datasource))
 
     query_dict = json.loads(dataset_result.datasetQuery)
-    if "query" in query_dict:
-        ds._query = DatasourceQuery.deserialize(query_dict["query"])
+
+    Datasource.deserialize_gql_result(ds, query_dict)
```
kbolashev, line 44:
```suggestion
    ds.deserialize_gql_result(query_dict)
```

Also probably should make it private by renaming it to `ds._deserialize_gql_result`, so users/IDEs won't think it's part of the public interface.

dagshub/data_engine/model/datasource.py
```diff
@@ -120,6 +154,17 @@ def all(self) -> "QueryResult":
         self._check_preprocess()
         return self._source.client.get_datapoints(self)
 
+    def select(self, *selected: Field):
+        """
+        Choose which columns will appear on the query result, what their names will be (alias) and from what time.
+        example:
+        t = int((datetime.datetime.now()-datetime.timedelta(hours=24)).timestamp())
+        q1 = (ds["episode"] > 5).select(Field("episode", as_of_time=t, alias="episode_asof_t"), Field("size"))
```
kbolashev, line 165: Python best practice: never use naive datetimes.
```suggestion
        Choose which columns will appear on the query result, what their names will be (alias) and from what time.
        example:
        from datetime import datetime, timedelta, timezone
        yesterday = datetime.now(timezone.utc) - timedelta(days=1)
        q1 = (ds["episode"] > 5).select(Field("episode", as_of_time=yesterday, alias="episode_yesterday"), Field("size"))
```

I'd go even as far as to say that we should explicitly not allow naive datetimes, because they can lead to unexpected results when ran on machines in different timezones.
You can read up a bit on it [here](https://realpython.com/python-datetime/#comparing-naive-and-aware-python-datetime-instances)

kbolashev: Also question regarding the whole select thing, backend included: If I apply a select, does it ONLY show the selected fields, or are they added on top of all the other basic fields?

Whatever the answer, it needs to be written out here

dagshub/data_engine/model/datasource.py
```diff
@@ -120,6 +154,17 @@ def all(self) -> "QueryResult":
...
+        """
+        self._select = [s.to_dict() for s in selected]
```
kbolashev: Looking at the docstring, having to do `Field("size")` looks clunky, I think we should be accepting strings here too:

```suggestion
        self._select = []
        for f in fields:
            if type(f) is str:
                self._select.append(Field(f).to_dict())
            else:
                self._select.append(f.to_dict())
```

kbolashev (changes requested):
I still don't have a strong opinion on the naive/aware timezone thing, whether we should block them completely, or try to figure out a workaround. Asking in the internal discord.

It's still good to have the docstring adhere by best practices.

dagshub/data_engine/model/datasource.py
```diff
@@ -91,13 +91,13 @@ def to_dict(self, ds: "Datasource") -> Dict[str, Any]:
 
 
 class Datasource:
-    def __init__(self, datasource: "DatasourceState", query: Optional[DatasourceQuery] = None):
+    def __init__(self, datasource: "DatasourceState", query: Optional[DatasourceQuery] = None, select=[], as_of=None):
```
kbolashev, line 94: Python annoying thing: you can't have mutable default values, because it's going to be a reference.
That means that if you append to the select, it might change values that you create later.
Better to have the select=None and in the constructor init it to [] if it's None

sdafni: oh yes, i should have known i ate shit from this in the past

docs/data_engine.md
```diff
@@ -281,6 +281,54 @@ filtered_ds = ds[ds["episode"] > 5]
...
+# size metadata is constantly changed and we want to address the one from 24h ago
+t = datetime.now(timezone.utc) - timedelta(hours=24)
+
+q1 = ds[Field("size", as_of_time=t] > 5
+```
+in the above example all datapoints whose "size" column updated no later than 't' that match the condition '>5' are returned.
+
+
+#### Query select:
+Using select() you can choose which columns will appear on the query result, what their names will be (alias) and from what time. For example:
```
kbolashev, line 297: Still don't know if these selected fields will apply **along** with the regular fields, or completely replacing them

docs/data_engine.md
```diff
@@ -281,6 +281,54 @@ filtered_ds = ds[ds["episode"] > 5]
...
+
+#### Notes and limitations:
+
+##### Time parameter:
+- the time parameter can be POSIX timestamp or datetime object
+- pay attention to timezones - use timestamp if known, or relative datetime if known (as in the above examples). if you use a specific date such as ```python dateutil.parser.parse("Tue 28 Nov 11:29 +2:00")``` specify the utc delta as shown here, otherwise this date can translate to different timestamps in the machine that runs the client and in dagshub backend.
+##### Select list:
+- both "x" and Field("x") can be used
+- alias, as_of_time - are optional
+- the list should make sense, i.e ```python.select(Field("x", as_of_time=t1), Field("x", as_of_time=t2))``` does not make sense since there is no alias to differentiate, the result will not reflect the intention. also ```python .select("x","x")```
```
kbolashev: "make sense" sounds kind of patronizing, especially since it's less about the fields actually not making sense, and more about us not accomodating for it.

Maybe reword it as something like "Right now if you select a single field multiple times, you need to give names to each of these fields"

docs/data_engine.md
```diff
@@ -281,6 +281,54 @@ filtered_ds = ds[ds["episode"] > 5]
...
+- when no select list specified all datapoint enrichements are returned, else only those specified.
```
kbolashev, line 325: Ok, this answered my question. This needs to be written at the very beginning of the function doc basically.

Also we 100% should have a flag to include the regular fields along with the selected ones.
Especially considering the way we write the docs, people might assume that the selected fields will be added to the original ones. That's how it read to me

docs/data_engine.md
```diff
@@ -281,6 +281,54 @@ filtered_ds = ds[ds["episode"] > 5]
...
+##### Global as_of behavior:
+- it applies to all entities unless otherwise specified, i.e if we use ```python Field("x", as_of_time=t1))``` than this t will take precedence over a t2 specified in .as_of(t2). the sensibility of the results is up to the caller. you could get datapoints that existed in t1 < t2 based on a condition applied on their enrichmnts in t2.
```
kbolashev: Also needs some rewording probably.

Fixing typos:
```suggestion
- it applies to all entities unless otherwise specified, i.e if we use ```python Field("x", as_of_time=t1))``` then this t will take precedence over a t2 specified in .as_of(t2). the sensibility of the results is up to the caller. you could get datapoints that existed in t1 < t2 based on a condition applied on their enrichments in t2.
```

Alternatively trying to reword it myself:
#### as_of specificity:
- `as_of` filters defined inside of `select()` take complete precedence over the `as_of` defined on top of the query. 

This means that a query:
```python
q = ds.select("size", Field("size", as_of_time=t1, alias="size_t1"), "train")
      .as_of(t2)
      .all()
```

Will return fields `size` and `train` as they were as of `t2`, and `size_t1` will have size as of `t1`.
This behavior is kept regardless of whether`t1` came before or after `t2`

kbolashev (approved):

PR #389, opened 2023-11-19
Add ability to download a repo folder (including CLI)

kbolashev:
This PR adds a capability to download a folder in the repository.
The `download` function is exposed in two places:
- On the `RepoAPI` object
- In the CLI as `dagshub download`

Caveats/possible improvements:
- Nonrecursive download doesn't create folder entries (only downloads files)
- The repo traversal is single-threaded


dagshub/common/cli.py
```diff
@@ -118,6 +119,48 @@ def to_log_level(verbosity):
         return logging.DEBUG
 
 
+@cli.command()
+@click.argument("repo", callback=validate_repo)
+@click.argument("remote_path")
+@click.argument("local_path", required=False, type=click.Path())
+@click.option("-b", "--branch", help="Branch or revision to download from")
```
guysmoilov, line 126:
```suggestion
@click.option("-b", "--branch", help="Branch or revision to download from. If left unspecified, use the default branch.")
```

dagshub/common/api/repo.py
```diff
@@ -226,6 +231,83 @@ def get_storage_file(self, path: str) -> bytes:
...
+
+    def download(
+        self, remote_path, local_path=".", revision=None, recursive=True, keep_source_prefix=True, redownload=False
+    ):
+        """
+        Downloads the contents of the repository at "remote_path" to the "local_path"
+
+        Args:
+            remote_path: path of the folder or folder to download in the repository
+            local_path: where to download the files. Defaults to CWD
```
guysmoilov, line 270:
```suggestion
            local_path: where to download the files. Defaults to current working directory.
```

dagshub/common/api/repo.py
```diff
@@ -226,6 +231,83 @@ def get_storage_file(self, path: str) -> bytes:
...
+            revision: repo revision, if not specified - uses default repo branch
+            recursive: whether to download files recursively
+            keep_source_prefix: whether to keep the path of the folder in the download path or not
+                example: Given remote_path "src/data" and file "test/file.txt"
+                if True: will download to "<local_path>/src/data/test/file.txt"
+                if False: will download to "<local_path>/test/file.txt"
+            redownload: whether to redownload an already existing file
+                The downloader doesn't do any hash comparisons and only checks if a file already exists in the FS or not
```
guysmoilov, line 278:
```suggestion
                The downloader doesn't do any hash comparisons and only checks if a file already exists in the local filesystem or not
```

dagshub/common/api/repo.py
```diff
@@ -226,6 +231,83 @@ def get_storage_file(self, path: str) -> bytes:
...
+        """
+
+        dir_queue = [path]
+        files = []
+
+        progress = get_rich_progress(rich.progress.MofNCompleteColumn())
+        task = progress.add_task("Traversing directories...", total=None)
+
+        def step(step_path):
+            res = self.list_path(step_path, revision)
```
guysmoilov, line 246: Will this also work with buckets?
If so, then:
1. It should be mentioned in the `branch` argument that revisions are ignored when downloading from buckets
2. The `:` in the bucket URL prefix needs to be handled correctly, and probably also documented

dagshub/common/api/repo.py
```diff
@@ -226,6 +231,83 @@ def get_storage_file(self, path: str) -> bytes:
...
+                The downloader doesn't do any hash comparisons and only checks if a file already exists in the FS or not
+        """
+        files = self._get_files_in_path(remote_path, revision, recursive)
+        file_tuples = []
+        if local_path is None:
+            local_path = "."
+        local_path = Path(local_path)
+        # Strip the slashes from the beginning so the relative_to logic works
+        if str(remote_path.startswith("/")):
+            remote_path = str(remote_path).lstrip("/")
```
guysmoilov, line 287: Can't you simplify?
Also, why the need to cast to str?
```suggestion
        remote_path = str(remote_path).lstrip("/")
```

dagshub/common/api/repo.py
```diff
@@ -226,6 +231,83 @@ def get_storage_file(self, path: str) -> bytes:
...
+        # Edge case - if the user requested a single file - different output path semantics
+        if len(files) == 1 and files[0].path == remote_path:
+            f = files[0]
+            remote_path = PurePosixPath(f.path)
+            # If local_path was specified, assume that the local_path is the exact name of the file
+            if local_path != Path("."):
+                file_path = local_path
```
guysmoilov, line 294: The local path might also be a directory, and to maintain the same semantic of `cp`, you should check that and if so copy the file to the dest dir without changing its name

guysmoilov (changes requested):

dagshub/common/cli.py
```diff
@@ -118,6 +119,48 @@ def to_log_level(verbosity):
...
+@click.option(
+    "--keep-prefix", is_flag=True, default=False, help="Whether to keep the path of the folder in the downloaded path"
+)
+@click.option("--not-recursive", is_flag=True, help="Don't download nested folders")
+@click.option("--redownload", is_flag=True, help="Redownload files, even if they already exist locally")
+@click.option("-v", "--verbose", default=0, count=True, help="Verbosity level")
+@click.option("-q", "--quiet", is_flag=True, help="Suppress print output")
```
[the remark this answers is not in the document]
kbolashev: Not in a way that would let them still show up in docs for this function, iirc.
I investigated that when we were integrating it, and that was an annoyance

dagshub/common/api/repo.py
kbolashev: Good point on simplification, not sure I need the if.
The str cast was defensive - in case remote_path is a Path-like, and not a string

dagshub/common/cli.py
```diff
@@ -118,6 +119,48 @@ def to_log_level(verbosity):
         return logging.DEBUG
 
 
+@cli.command()
+@click.argument("repo", callback=validate_repo)
+@click.argument("remote_path")
+@click.argument("local_path", required=False, type=click.Path())
+@click.option("-b", "--branch", help="Branch or revision to download from")
+@click.option(
+    "--keep-prefix", is_flag=True, default=False, help="Whether to keep the path of the folder in the downloaded path"
```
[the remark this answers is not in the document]
kbolashev: I'm afraid of dumping man-file level of verbosity onto a user here. I can copy the explanation from the function itself I guess if you want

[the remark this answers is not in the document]
guysmoilov: IDK, this is the human-facing explanation. Almost no one opens source code.

dagshub/common/api/repo.py
kbolashev: The buckets should be working almost out of the box, with the exception of the `:` symbol not being valid in filenames in windows, so you're right that we'd need to scrub it.

There is a caveat though, and a thing that I should change and it's not hard - the bucket downloader optimization only works on paths that go through `storage/s3/bucket`, and not `src/s3:/bucket`. But that should be easy for me to change from latter to former.

dagshub/common/cli.py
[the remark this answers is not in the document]
kbolashev: valid

dagshub/common/api/repo.py
kbolashev: Handling the buckets correctly now.
I even managed to do a `repo.download("/", "local/path")` and it even downloaded E V E R Y T H I N G

Let me know if you maybe want me to put auto-downloading storages behind a flag?

dagshub/common/api/repo.py
kbolashev: Did it so if we're downloading to an existing directory, then it's doing this.
Not sure how to handle something like that with a directory that doesn't exist _yet_, so say something like
```python
repo.download("src/file.py", "/non/existent/dir")
```

Not sure if it's worth trying to come up with a workable semantic there, because everyone will have their own weird edge cases. Hidden directories, filenames without extensions, dotfiles.
But then now we're behaving differently depending on whether the folder exists or not, so calling 
```python
repo.download("src/file.py", "/non/existent/dir")
```
on a machine where `/non/existent/dir` exists will save it to `/non/existent/dir/file.py`
and on a machine where it doesn't, it will save the file to `/non/existent/dir` (the file named "dir")

Also how would you expect the remote prefix to work here? Should it save to `dir/src/file.py` or `dir/file.py`?

dagshub/common/api/repo.py
guysmoilov: @kbolashev Great, yeah I think there should be a flag

dagshub/common/api/repo.py
guysmoilov: 1. About missing directories - we should emulate the semantics of `cp -r` - so if you're attempting to download a file to a "non-existent-directory", that would mean that you should rename the file to the name of the non-existent directory. If you're downloading a directory, it should rename that directory to the non-existent directory. If you have a non-existent directory in the middle of the target path rather than its edge, then the whole copy command should throw an error (unless maybe a `mkdirs` flag is set to True)
2. If preserving the prefix, it means you assume that the target path is always a directory and you create your tree structure inside of it

dagshub/common/api/repo.py
```diff
@@ -259,78 +260,134 @@ def get_storage_file(self, path: str) -> bytes:
...
     def download(
-        self, remote_path, local_path=".", revision=None, recursive=True, keep_source_prefix=True, redownload=False
+        self,
+        remote_path,
+        local_path=".",
+        revision=None,
+        recursive=True,
+        keep_source_prefix=False,
+        redownload=False,
+        download_storages=False,
```
guysmoilov, line 326: I think you forgot to expose this in the CLI

dagshub/common/api/repo.py
```diff
@@ -259,78 +260,134 @@ def get_storage_file(self, path: str) -> bytes:
...
         local_path = Path(local_path)
         # Strip the slashes from the beginning so the relative_to logic works
-        if str(remote_path.startswith("/")):
-            remote_path = str(remote_path).lstrip("/")
+        remote_path = str(remote_path).lstrip("/")
+        if not remote_path:
+            remote_path = "/"
+        # For storage paths get rid of the colon in the beginning of the schema, the download urls won't have it either
+        if remote_path.split("/")[0] in {"s3:", "gs:", "azure:"}:
+            remote_path = remote_path.replace(":", "", 1)
```
guysmoilov, line 367: Note that this replaces potential colons in the path, not just the scheme

dagshub/common/api/repo.py
```diff
@@ -259,78 +260,134 @@ def get_storage_file(self, path: str) -> bytes:
...
         # Edge case - if the user requested a single file - different output path semantics
         if len(files) == 1 and files[0].path == remote_path:
             f = files[0]
             remote_path = PurePosixPath(f.path)
             # If local_path was specified, assume that the local_path is the exact name of the file
             if local_path != Path("."):
-                file_path = local_path
+                # Saving to existing dir - append the name of remote file to the end a-la cp
+                if local_path.exists() and local_path.is_dir():
```
guysmoilov, line 375: If it exists and not a dir, you should throw.
Also, make sure this works with symlinks (in the middle and end of the paths)

dagshub/common/api/repo.py
```diff
@@ -259,78 +260,134 @@ def get_storage_file(self, path: str) -> bytes:
...
             local_path = "."
         local_path = Path(local_path)
         # Strip the slashes from the beginning so the relative_to logic works
-        if str(remote_path.startswith("/")):
-            remote_path = str(remote_path).lstrip("/")
+        remote_path = str(remote_path).lstrip("/")
+        if not remote_path:
+            remote_path = "/"
+        # For storage paths get rid of the colon in the beginning of the schema, the download urls won't have it either
+        if remote_path.split("/")[0] in {"s3:", "gs:", "azure:"}:
```
guysmoilov, line 366: I see that this check is repeated also in the _get_files_in_path function, with a slight modification of the lstrip which you might also want here?
Maybe you should refactor this into a reusable function?

dagshub/common/api/repo.py
```diff
@@ -259,78 +260,134 @@ def get_storage_file(self, path: str) -> bytes:
...
+
+        def push_folder(folder_path):
+            dir_queue.append((folder_path, list_fn_folder))
+
+        def push_storage(storage_path):
+            dir_queue.append((storage_path, list_fn_storage))
+
+        # Initialize the queue
+        is_storage_path = str(path).lstrip("/").split("/")[0] in {"s3:", "gs:", "azure:"}
+        if is_storage_path:
```
guysmoilov, line 284: If the root is a storage path, I think you have to set `traverse_storages=True` to correctly handle the recursion

guysmoilov (changes requested):

dagshub/common/api/repo.py
kbolashev: Right, thank you, totally forgot

dagshub/common/api/repo.py
kbolashev: The 1 at the end of the `remote_path.replace()` is how many times to change the value, so that way we're only changing the scheme (we checked that it started with it anyway)

dagshub/common/api/repo.py
kbolashev: `traverse_storages` is False ONLY when the root is being traversed (in the caller of this function).

Basically I wanted to make it so users don't need to explicitly set `traverse_storages=True` when downloading `s3://my-bucket`, because I think that'd be a surprising behavior for them (why do i have to flip this flag when I'm downloading a bucket folder but not a repo folder)

dagshub/common/api/repo.py
kbolashev: If it exists and not a dir, it's going to defer to the `redownload` flag. If it exists, it just won't be downloaded

kbolashev: I don't think I ever had problems with pathlib and symlinks

guysmoilov (approved):

PR #393, opened 2023-11-30
dict update #1

sdafni:
https://discord.com/channels/696726028751732806/1176106633705828473/1176106637329694750

implement #1 from the above:

datapoint[column] = value

TODO: #2

### DIFF
--- a/dagshub/data_engine/model/datasource.py
+++ b/dagshub/data_engine/model/datasource.py
@@ -127,11 +127,15 @@ class Datasource:
         self._select = select or []
         self._global_as_of = as_of
         self.serialize_gql_query_input()
-        self._ds_scope_update_ctx = None
+        self._implicit_update_ctx = None
+        self.explicit_update_ctx_counter = 0
+
+    def has_context(self):
+        return self.explicit_update_ctx_counter > 0
 
     def _get_source_scope_metadata_entries(self):
-        if self._ds_scope_update_ctx is not None:
-            return self._ds_scope_update_ctx.get_metadata_entries()
+        if self._implicit_update_ctx is not None:
+            return self._implicit_update_ctx.get_metadata_entries()
         else:
             return []
 
@@ -335,18 +339,18 @@ class Datasource:
         self.source.get_from_dagshub()
 
     def get_ctx(self) -> "MetadataContextManager":
-        if not self._ds_scope_update_ctx:
+        if not self._implicit_update_ctx:
             self.source.get_from_dagshub()
-            self._ds_scope_update_ctx = MetadataContextManager(self)
+            self._implicit_update_ctx = MetadataContextManager(self)
 
-        return self._ds_scope_update_ctx
+        return self._implicit_update_ctx
 
     def save_ctx(self):
-        if self._ds_scope_update_ctx:
+        if self._implicit_update_ctx:
             try:
                 self._upload_metadata(self._get_source_scope_metadata_entries())
             finally:
-                self._ds_scope_update_ctx = None
+                self._implicit_update_ctx = None
 
     def metadata_context(self) -> ContextManager["MetadataContextManager"]:
         """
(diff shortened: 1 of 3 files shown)


dagshub/data_engine/model/datasource.py
```diff
@@ -344,8 +372,11 @@ def func():
             self.source.get_from_dagshub()
             send_analytics_event("Client_DataEngine_addEnrichments", repo=self.source.repoApi)
             ctx = MetadataContextManager(self)
+
+            ctx.start_explicit_ds_ctx()
```
kbolashev, line 376: I think it's better to just store it in another field in the datasource object

```
ds._explicit_update_context = ctx
```

And you set it to None after you're done with it.

kbolashev: That'd skip having to do the indirection with the counter

dagshub/data_engine/model/datasource.py
```diff
@@ -344,8 +372,11 @@ def func():
...
             yield ctx
-            self._upload_metadata(ctx.get_metadata_entries())
+            self._upload_metadata(ctx.get_metadata_entries() + self._get_source_implicit_metadata_entries())
```
kbolashev, line 378: You should empty the implicit context on upload finish (maybe in a try/finally block).
Otherwise the values updated in the implicit context will be uploaded again next time you do another `with ds.metadata_context()`.
I guess you can actually repeat the code you did in `save_ctx`, setting the implicit context to None will also work

dagshub/data_engine/model/datasource.py
```diff
@@ -327,6 +341,20 @@ def apply_field_changes(self, field_builders: List[MetadataFieldBuilder]):
...
         self.source.get_from_dagshub()
 
+    def get_ctx(self) -> "MetadataContextManager":
+        if not self._implicit_update_ctx:
+            self.source.get_from_dagshub()
+            self._implicit_update_ctx = MetadataContextManager(self)
+
+        return self._implicit_update_ctx
+
+    def save_ctx(self):
```
kbolashev, line 351: Please name all of the functions related to metadata updating with `_update_metadata_` in them.
Also add docstrings on all of the functions explaining what they're doing.

Add a `:meta private:` line if it shouldn't be showing up in the docs (which are all functions in this PR I think, aside from `Datapoint.save()`)

dagshub/data_engine/model/datasource.py
```diff
@@ -127,6 +127,20 @@ def __init__(self, datasource: "DatasourceState", query: Optional[DatasourceQuer
...
         self._global_as_of = as_of
         self.serialize_gql_query_input()
+        # a per datasource context used for dict update syntax
+        self._implicit_update_ctx = None
+        # this counter marks if source is currently used in
+        # meta-data update 'with' block
+        self.explicit_update_ctx_counter = 0
+
+    def has_context(self):
+        return self.explicit_update_ctx_counter > 0
```
kbolashev, line 137:
```suggestion
    @property
    def has_explicit_context(self):
        return self._explicit_update_ctx is not None
```

kbolashev (changes requested):

dagshub/data_engine/model/datasource.py
sdafni: which functions to rename ?  currently nothing is named _update_metadata_XXX

dagshub/data_engine/model/datasource.py
```diff
@@ -327,6 +342,24 @@ def apply_field_changes(self, field_builders: List[MetadataFieldBuilder]):
...
+    def implicit_update_context(self) -> "MetadataContextManager":
+        if not self._implicit_update_ctx:
+            self._implicit_update_ctx = MetadataContextManager(self)
+
+        return self._implicit_update_ctx
+
+    def save_ctx(self):
+        """
+        commit meta data changes done in dictionary assignment context
+        :meta private
```
kbolashev, line 355:
```suggestion
        :meta private:
```

dagshub/data_engine/model/datasource.py
kbolashev: That's my point, `save_ctx` doesn't mean much, and since the Datasource object is responsible for a lot of stuff, you need to have functions be named relating to what they're doing.
So with that in mind, this should be named something like `save_implicit_update_ctx`

kbolashev (approved):
Fix the `save_ctx` naming stuff + docs and you can merge

PR #401, opened 2023-12-13
include/ exclude list for LS meta-data

sdafni:
TODO needs documentation

### DIFF
--- a/dagshub/data_engine/model/query_result.py
+++ b/dagshub/data_engine/model/query_result.py
@@ -513,23 +513,23 @@ class QueryResult:
 
     def annotate(self, open_project=True,
                  ignore_warning=True,
-                 ls_meta_excludes=None,
-                 ls_meta_includes=None) -> Optional[str]:
+                 fields_to_exclude=None,
+                 fields_to_embed=None) -> Optional[str]:
         """
         Sends all the datapoints returned in this QueryResult to be annotated in Label Studio on DagsHub.
 
         Args:
             open_project: Automatically open the Label Studio project in the browser
             ignore_warning: Suppress the prompt-warning if you try to annotate too many datapoints at once.
-            ls_meta_includes: list of meta-data columns that will show up in Label Studio UI.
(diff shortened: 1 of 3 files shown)


dagshub/data_engine/model/query_result.py
```diff
@@ -511,19 +511,25 @@ def visualize(self, **kwargs):
 
         return sess
 
-    def annotate(self, open_project=True, ignore_warning=True) -> Optional[str]:
+    def annotate(self, open_project=True,
+                 ignore_warning=True,
+                 ls_meta_excludes=None,
+                 ls_meta_includes=None) -> Optional[str]:
```
kbolashev, line 517: Need to add the new kwargs to `Datasource.annotate()` also

kbolashev (changes requested):
Don't like the `ls_meta_includes/excludes` naming. Very programmer-y and requires you to know what "ls" and "meta" stands for.

I don't have a good idea for an alternative name in mind for now. Something like `fields_to_embed` maybe. I'll try to come up with some variants and ask on Discord.

Aside from that, the UX is also very unfriendly imo.
Here's my proposition:

- 1 arg instead of 2
- Accept either a string or a list.
- If it's a string, then turn it into a list by `arg.split(",")`.
- `"field"` goes into the includes.
- `"!field"` goes into the excludes.

I checked and we don't allow `,` and `!` in field names, so we should be clear on this
The only problem I see with the string approach is we allow field names to begin/ends with spaces, which is a mistake. Need to talk it out with Simon and fix it maybe :)

kbolashev:
Also a nightmare scenario, that probably needs to be discussed if we want to support it, from the glance at the backend code it doesn't look like it's going to work.

What happens if I do:

```python
qr = ds.select(Field(size, as_of=yesterday, alias="size_yesterday")).all()
qr.annotate(ls_meta_includes="size_yesterday")
```

kbolashev (approved):
Thank you!!

PR #402, opened 2023-12-17
Get and Save Datasets and Datasources to Files

deanp70:
Add functionality to save and get datasets and datasources to files

### DIFF
--- a/dagshub/data_engine/client/models.py
+++ b/dagshub/data_engine/client/models.py
@@ -89,3 +89,16 @@ class DatasetResult:
     name: str
     datasource: DatasourceResult
     datasetQuery: str
+
+
+@dataclass_json
+@dataclass
+class DatasourceFileStruct:
+    id: Union[str, int]
+    repo: str
+    name: str
+    datasetQuery: Optional[dict]
+
+    @property
+    def has_query(self):
+        return (self.datasetQuery is not None) and (self.datasetQuery["query"] is not None)
--- a/dagshub/data_engine/client/models.py
+++ b/dagshub/data_engine/client/models.py
@@ -101,4 +101,4 @@ class DatasourceFileStruct:
 
     @property
     def has_query(self):
-        return (self.datasetQuery is not None) and (self.datasetQuery["query"] is not None)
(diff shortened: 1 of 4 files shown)


dagshub/data_engine/client/models.py
```diff
@@ -89,3 +89,16 @@ class DatasetResult:
...
+@dataclass
+class DatasourceFileStruct:
+    id: Union[str, int]
+    repo: str
+    name: str
+    datasetQuery: Optional[dict]
+
+    @property
+    def has_query(self):
+        return (self.datasetQuery is not None) and (self.datasetQuery["query"] is not None)
```
kbolashev, line 104: more defensive
```suggestion
        return (self.datasetQuery is not None) and (self.datasetQuery.get("query") is not None)
```

dagshub/data_engine/model/datasource.py
```diff
@@ -529,6 +530,26 @@ def save_dataset(self, name: str):
         self.source.client.save_dataset(self, name)
         log_message(f"Dataset {name} saved")
 
+    def save_to_file(self, path: str = ".", name: str = ""):
```
kbolashev, line 533: Could be useful to return the path to the file as a result of this function, so a Python script can do something else with the resulting file immediately (push it to dagshub, or w/e the user would want)

dagshub/data_engine/datasets.py
```diff
@@ -45,6 +46,19 @@ def get_dataset(repo: str, name: Optional[str] = None, id: Optional[Union[int, s
     return sources[-1]
 
 
+def get_dataset_from_file(path: str):
+    """
+    Load a dataset from a local file
+
+    Args:
+        path: Path to the `.dagshub` file with the relevant dataset
```
kbolashev, line 54: The sphinx syntax requires double backticks for the code blocks

```suggestion
    Args:
        path: Path to the ``.dagshub`` file with the relevant dataset
```

kbolashev (approved):
Looks awesome! Thank you!
Adding some nice-to-haves

dagshub/data_engine/model/datasource.py
deanp70: Done

PR #409, opened 2023-12-20
Bugfix: Fixed a faulty null check

jinensetpal:


dagshub/data_engine/model/datasource_state.py
```diff
@@ -103,7 +103,7 @@ def raw_path(self, path: Union[str, Datapoint, Mapping[str, Any]]) -> str:
     @property
     def source_prefix(self) -> PurePosixPath:
         parts = self.path_parts()
-        if "prefix" in parts:
+        if parts["prefix"]:
```
kbolashev, line 106:
```suggestion
        if parts.get("prefix"):
```

kbolashev (changes requested):
Need to still check IF "prefix" exists there, but I didn't think that the value itself could be None, so thanks for catching it.

PR #410, opened 2023-12-21
Save/Load Datasource/Dataset state from MLflow

kbolashev:
Scope of changes:

[Prerequisites to this PR's features]
- Persist information about the dataset in the `Datasource` object
- Add new fields to `DatasourceSerializedState` object (renamed from `DatasourceFileStruct`): info about the dataset, save timestamp, link, was the query modified from the dataset's query

[Additional stuff implemented along the way]
- `save_dataset()` now sets the dataset as the one assigned to the datasource. Togglable with `assign_after_save=False`
- `clear_query()` resets the query to the dataset's one by default. Togglable with `reset_to_dataset=False`
- `dagshub.init()` with MLflow now explicitly logs which repo we're now tracking mlflow into.
- Information about the dataset is printed out in the `ds.__repr__`, so it will show up when printing it out in e.g. Jupyter cell.
- `black` formatting pass over the datasource querying test (sorry for the huge LOC change over there)

[Feature requests of this PR]
- Add `ds.log_to_mlflow()` function, saving the state of the datasource into a json file, uploading that as an artifact to the MLflow's current run.
- Add `datasources.get_from_mlflow(run_id, artifact_name)` and `datasets.get_from_mlflow(run_id, artifcat_name)` functions that allow importing this serialized state back into python.


dagshub/data_engine/model/datapoint.py
```diff
@@ -47,10 +47,20 @@ def __setitem__(self, key, value):
...
+        Commit metadata changes on this datapoint.
+
+        Example::
+
+            datapoint["field"] = value
+            datapoint.save()
+
+        .. note::
+            If done inside a metadata context, ``save()`` doesn't need to be called,
+            all changes will be uploaded on context exit::
```
guysmoilov, line 59:
```suggestion
            If done inside a metadata context, ``save()`` doesn't need to be called,
            all changes will be uploaded on context exit or in efficient batches.
            It's recommended to use the metadata context when editing a large amount of datapoints::
```

dagshub/data_engine/model/datasource.py
```diff
@@ -120,7 +129,14 @@ def to_dict(self, ds: "Datasource") -> Dict[str, Any]:
 
 
 class Datasource:
-    def __init__(self, datasource: "DatasourceState", query: Optional[DatasourceQuery] = None, select=None, as_of=None):
+    def __init__(
+        self,
+        datasource: "DatasourceState",
+        query: Optional[DatasourceQuery] = None,
+        select: Optional[List[Dict]] = None,
```
guysmoilov, line 136: Shouldn't the `Dict` be typed?

dagshub/data_engine/model/datasource.py
```diff
@@ -158,18 +175,26 @@ def global_as_of_timestamp(self) -> Optional[int]:
...
+            self._query = self.assigned_dataset.query.__deepcopy__()
+        else:
+            self._query = DatasourceQuery()
 
     def __deepcopy__(self, memodict={}) -> "Datasource":
         res = Datasource(self._source, self._query.__deepcopy__(), self._select, self._global_as_of)
+        res.assigned_dataset = self.assigned_dataset
+
+        # Carry over the update context, that way we'll keep track of the metadata being uploaded
+        res._implicit_update_ctx = self._implicit_update_ctx
```
guysmoilov, line 196: What about the explicit context?

dagshub/data_engine/model/datapoint.py
```diff
@@ -47,10 +47,20 @@ def __setitem__(self, key, value):
...
+
+                with ds.metadata_context():
+                    for i, dp in ds.all():
+                        dp["number"] = i
         """
 
         # if in context block, don't _upload_metadata, it will be done at context end
```
guysmoilov, line 66: I think there's a subtle bug here:
1. Get datapoint 1 from a datasource
2. Modify it - implicit upload context gets created
3. Get datapoint 2
4. Modify it - keeps existing context, adds to it
5. Save datapoint 1 - it actually saves datapoint 2 as well!

guysmoilov: I can live with it but it's bad ACID or something

dagshub/data_engine/model/datasource.py
```diff
@@ -543,49 +617,111 @@ def _upload_metadata(self, metadata_entries: List[DatapointMetadataUpdateEntry])
...
+        res = self._serialize().to_dict()
+        # Skip Nones in the result
+        res = {k: v for k, v in res.items() if v is not None}
+        return res
+
+    @property
+    def is_query_different_from_dataset(self) -> Optional[bool]:
+        if self.assigned_dataset is None:
+            return None
+        return self._query.to_dict() != self.assigned_dataset.query.to_dict()
```
guysmoilov, line 700: Will this comparison actually work?
Seems like you need a deep comparison

dagshub/data_engine/model/datasource.py
```diff
@@ -543,49 +617,111 @@ def _upload_metadata(self, metadata_entries: List[DatapointMetadataUpdateEntry])
...
         .. note::
-            Does not save dataset contents, only a pointer file.
+            Does not save dataset contents.
 
         Args:
-            path: Path to save the datasource or dataset file in, defaults to current directory
-            name: Optional: name of the file. if not provided, datasource or dataset name will be used instead.
+            path: Path to the directory where to save the datasource or dataset file.
+                Defaults to current working directory.
+            name: Optional: name of the file. if not provided, datasource name will be used instead.
```
guysmoilov, line 663: It should use the dataset name if it's assigned

dagshub/data_engine/model/datasource.py
```diff
@@ -543,49 +617,111 @@ def _upload_metadata(self, metadata_entries: List[DatapointMetadataUpdateEntry])
...
-            repo=self._source.repo,
-            name=self._source.name,
-            datasetQuery=self.serialize_gql_query_input()
-        )
-        file_path = os.path.join(path, (name or res.name) + ".dagshub")
+        file_path = os.path.join(path, (name or self.source.name) + ".json")
+        res = self._to_dict()
         with open(file_path, "w") as file:
-            file.write(json.dumps(res.to_dict(), indent=4))
+            file.write(json.dumps(res, indent=4))
```
guysmoilov, line 671: I'm pretty sure you can write directly to the file + sort by keys maybe, for consistent ordering?
Or else, use an ordered dict.
Or does your dataclass output ordered dicts by default?
```suggestion
            json.dump(res, file, indent=4, sort_keys=True)
```

dagshub/data_engine/model/datasource.py
```diff
@@ -543,49 +617,111 @@ def _upload_metadata(self, metadata_entries: List[DatapointMetadataUpdateEntry])
...
+        ds_state.get_from_dagshub()
+        ds = Datasource(ds_state)
+
+        if state.has_query:
+            ds._deserialize_from_gql_result(state.query)
+
+        if state.dataset_id is not None:
+            ds.load_from_dataset(state.dataset_id, state.dataset_name, change_query=False)
+
+        return ds
```
guysmoilov, line 723: I think it's pretty clumsy that you need to use both the `state` and the `ds_state` together to construct one `Datasource` instance, don't you think?
When you pass `ds_state` into the constructor, I kind of expect it to do everything needed.

I might just be misunderstanding the point of `DatasourceState` in general - why does it exist?
Is it a low level object intended for API communication, while the `Datasource` object is for a convenient UX for users?
If so, their names are kind of confusing. Maybe I would call the `DatasourceState` something like `DatasourceClient` or Backend or Dto or something.

guysmoilov, line 723: I think it's pretty clumsy that you need to use both the `state` and the `ds_state` together to construct one `Datasource` instance, don't you think?
When you pass `ds_state` into the constructor, I kind of expect it to do everything needed

guysmoilov (changes requested):
Partial review

dagshub/data_engine/client/data_client.py
```diff
@@ -71,7 +71,7 @@ def create_datasource(self, ds: "DatasourceState") -> DatasourceResult:
         res = self._exec(q, params)
         return dacite.from_dict(DatasourceResult, res["createDatasource"], config=dacite_config)
 
-    def head(self, datasource: Datasource, size: Optional[int] = None) -> QueryResult:
+    def head(self, datasource: "Datasource", size: Optional[int] = None) -> QueryResult:
```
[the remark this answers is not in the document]
kbolashev: There was a circular import with the DatasourceSerializedState. I wanted to move deserialization into Datasource, but one of the dependencies along the way ended up importing the data client.
When I switch to the string type-hinting I hide the import behind an `if TYPE_CHECKING`, but at runtime Datasource no longer gets imported, eleminating the circular import.

dagshub/data_engine/client/models.py
```diff
@@ -93,12 +93,34 @@ class DatasetResult:
...
     repo: str
-    name: str
-    datasetQuery: Optional[dict]
+    """Repository this datasource is on"""
+    datasource_id: Union[str, int]
+    """ID of the datasource"""
+    datasource_name: str
+    """Name of the datasource"""
+    query: Optional[dict] = None
+    """Query at the time of saving"""
```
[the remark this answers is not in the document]
kbolashev: Not necessary imo, but we could do it. Would be a `QuerySerializedState` that can be cast to `DatasourceQuery`

dagshub/data_engine/client/models.py
```diff
@@ -93,12 +93,34 @@ class DatasetResult:
 
 @dataclass_json
```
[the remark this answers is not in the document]
kbolashev: I'm actually mainly using this library to get a `to_dict()` function on a dataclass. The actual JSON printing is, for us, controlled by the `json.dumps()`.
I think the library has some additional json sugar, but i didn't investigate it further yet

dagshub/data_engine/model/dataset_state.py
```diff
@@ -0,0 +1,62 @@
+import json
+from dataclasses import dataclass
+from typing import Dict, Optional, List, Union
+
+from dagshub.data_engine.client.models import DatasetResult
+from dagshub.data_engine.model.query import DatasourceQuery
+
+
+@dataclass
+class DatasetState:
```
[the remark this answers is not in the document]
kbolashev: This one is also cached inside of `Datasource` (`ds.assigned_dataset`), so datasource can know the query of the dataset and other info relating to that.
I think it'd be weird to have a `DatasourceSerializedState` be a field in `Datasource`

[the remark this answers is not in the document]
kbolashev: Also the Query in this one is not really serializable.
I guess I can make the DatasourceQuery object serializable/deserializable by itself. That'd probably simplify things a lot 🤔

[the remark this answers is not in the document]
kbolashev: will cut a couple serialization sub-structs

dagshub/data_engine/model/datasource.py
kbolashev: the typing is just a suggestion, it's a str/dict dict so it won't really be helpful

dagshub/data_engine/model/datasource.py
kbolashev: that one doesn't really matter because it only gets uploaded once you exit the context.
We're not losing the reference to it, so the explicit context is persisting.

So examples for code where the implicit context passing is relevant

Implicit context passing (doesn't work currently):
```python
dps = ds.all()
dp = dps[0]
dp["field1"] = 2

queried = ds["sub_query"] == "fsadfasfd"
dp2 = queried.all()[0]
dp2["field2"] = 3

dp2.save()  # Will save both the dp and the dp2 change
```

Explicit context passing:
```python
with ds.update_metadata() as ctx:
  ctx.update_metadata(...)
  queried = ds[..] == ...
  with queried.update_metadata() as ctx2:
    ctx2.update_metadata(...)
  # Uploads ctx2
#Uploads ctx
...
```

dagshub/data_engine/model/datapoint.py
kbolashev: yes, you're right, it totally works that way.
I guess it's a bit of an underexplained behavior? You'd probably expect `dp.save()` to upload only that datapoint's changes.
I'll think about changing it maybe

dagshub/data_engine/model/datasource.py
kbolashev: I checked and it actually works. A couple of the tests rely on that.
Works because you only need deep comparison if the subelements in the dict are objects. `query.to_dict()` returns dict with only primitive keys and values, so it's fine.

dagshub/data_engine/model/datasource.py
kbolashev: good idea, will change that

dagshub/data_engine/model/datasource.py
kbolashev: > I might just be misunderstanding the point of DatasourceState in general - why does it exist?
Is it a low level object intended for API communication, while the Datasource object is for a convenient UX for users?

Yes, you're on the money here.
DatasourceState keeps the RepoAPI object and the DataClient object, plus additionally has helper functions to generate paths/download urls.
`Datasource` is more of the user-facing side, and it also keeps info about the query (and the dataset now). Meanwhile`DatasourceState` is the part that ties it to DagsHub and the part that's aware of the repo it's tied to.

I expected that `Datasource` would get big, so that was the boundary I set. We can totally rename it, since it's not really part of the public interface of the data engine.

dagshub/data_engine/model/datapoint.py
kbolashev: Would require to separate the dict assignment inside of a context from dict assignment that's coupled with `save()`. Or maybe would require to delete it from the context somehow.

dagshub/data_engine/model/dataset_state.py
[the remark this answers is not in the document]
kbolashev: Took a glance, I'll have to make `DatasourceQuery` a dataclass, which it isn't right now, which might be going into abuse territory, because there's also the whole hidden state with the tree, which i DON'T want to be serializing/deserializing.

Maybe I should write a custom [JSONEncoder](https://docs.python.org/3/library/json.html#json.JSONEncoder) and use that throughout Data Engine files, but that interface is pretty clunky imo

dagshub/data_engine/model/datasource.py
kbolashev: ds_state lacks any information about the dataset/query, that's why the constructor has the additional fields aside from `DatasourceState`. Meanwhile `DatasourceSerializedState` has both the Datasource information and Query information.

I guess this goes hand in hand with your complaint about the actual purpose of the `DatasourceState` class. Not sure what I should be doing about it though?

kbolashev:
Mostly doc/style changes, didn't change anything functionally/no class shuffles. I tried to explain my reasoning for everything, but let me know if you're unconvinced and want me to change stuff.

Honestly right now the only thing I think _should_ be done is moving the `as_of` and `select` from `Datasource` into `DatasoureQuery`, since they are much more a part of the query. And I think I should try to figure out how to make `DatasourceQuery` more easier to serialize/deserialize to json

guysmoilov (changes requested):

dagshub/data_engine/model/datapoint.py
kbolashev: But if we don't have an implicit context, we're throwing away the functionality that we're commenting on here (the snippet in the comment).
I think it's neat and is very convenient without really any drawbacks.
I can remove it if you really want to, but i think it's good UX.

kbolashev: I guess we can still have this implicit context behavior by passing the context into `save()`, with throwaway context with immediate upload by default.

dagshub/data_engine/model/datasource.py
kbolashev: I think you're right and it might not work with the nested query? I'll make sure to check, thank you.

dagshub/data_engine/model/datasource.py
```diff
@@ -543,49 +617,122 @@ def _upload_metadata(self, metadata_entries: List[DatapointMetadataUpdateEntry])
...
             The path to the saved file
         """
-        res = DatasourceFileStruct(
-            id=self._source.id,
-            repo=self._source.repo,
-            name=self._source.name,
-            datasetQuery=self.serialize_gql_query_input()
-        )
-        file_path = os.path.join(path, (name or res.name) + ".dagshub")
+        if not name:
```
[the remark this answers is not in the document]
kbolashev: You're totally right, didn't even come to my mind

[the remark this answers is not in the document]
kbolashev: Changed it :)

dagshub/data_engine/model/datapoint.py
kbolashev: Would you be ok with moving this to another task then? This is kind of out of scope of this PR at this point

dagshub/data_engine/client/data_client.py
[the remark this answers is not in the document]
guysmoilov: I wasn't aware that circular imports are an issue in Python, but OK

dagshub/data_engine/model/dataset_state.py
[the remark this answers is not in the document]
guysmoilov: I'm a bit confused on how the query isn't serializeable - you obviously do serialize it.
Not that any of it is critical for merging the PR...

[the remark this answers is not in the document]
kbolashev: It is de/serializable, but only in the context of "I need to call a whole another function to actually do that". I don't know of a neat way to change the signature in DatasourceSerializedState to DatasourceQuery and say "hey actually you need to use this function to serialize, and this to deserialize"

kbolashev:
Rolled `as_of` and `select` into `DatasourceQuery`, change serialization/deserialization, so `DatasourceSerializedState.query` is now `DatasourceQuery`

dagshub/data_engine/model/dataset_state.py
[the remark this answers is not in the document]
kbolashev: Made `DatasourceSerializedState` actually have the `DatasourceQuery` in it in the end :)

dagshub/data_engine/model/datapoint.py
guysmoilov: Sure, no problem doing this in a separate PR

dagshub/data_engine/model/query.py
```diff
@@ -54,20 +54,30 @@ class FieldFilterOperand(enum.Enum):
...
 
 class DatasourceQuery:
-    def __init__(self, column_or_query: Optional[Union[str, "DatasourceQuery"]] = None, as_of: Optional[int] = None):
+    def __init__(
+        self,
+        column_or_query: Optional[Union[str, "DatasourceQuery"]] = None,
+        field_as_of: Optional[int] = None,
+        query_as_of: Optional[int] = None,
+        selects: Optional[List] = None,
+    ):
```
guysmoilov, line 63: Seems to me like there's some type confusion here between top level query and the leaf / intermediate node query objects:
1. query_as_of should only belong on the top level, why pass it into all the internal nodes?
2. Same for selects
3. I'll have a related comment on the `__getitem__` of the Datasource

guysmoilov: Meaning, the top level query is probably best defined as its own class...
Or at most, an explicit subclass of DatasourceQuery with additional fields, rather than the current thing which is like a fragile implicit subclass

dagshub/data_engine/model/query.py
```diff
@@ -145,6 +155,10 @@ def compose(self, op: str, other: Optional[Union[str, int, float, "DatasourceQue
             composite_tree.paste(root_node.identifier, self._operand_tree)
             composite_tree.paste(root_node.identifier, other._operand_tree)
             self._operand_tree = composite_tree
+            if other.query_as_of is not None:
+                self.query_as_of = other.query_as_of
+            if other.selects is not None:
+                self.selects = other.selects
```
guysmoilov, line 161: Unclear to me why you'd want to copy this around to the internal nodes, as noted earlier.

dagshub/data_engine/model/query.py
guysmoilov: I think I'm confused since the whole class hierarchy doesn't make sense to me, in my head (with intuition from my knowledge of Pandas), when doing `ds["bla"]` a Query should be returned, not a Datasource/set.
Something like `ds["bla"].all()` will only make sense if it returns an array of the "bla" column instead of the whole dataset dataframe...
I think that's the root of my confusion, and that might mean that my review here and/or the change that pushes query_as_of and selects into the Query class doesn't make much sense anyway

guysmoilov (changes requested):

dagshub/data_engine/model/query.py
kbolashev: the new query completely replaces the old one. If you compose a couple queries but don't carry it over, you'll lose the select/as of

Example:
```python
q1 = (ds["data"] == "5").as_of(t1)
q2 = (ds["other_data"] > 5).select("data", "other_data")
q3 = q1 & q2
```
you'd probably expect q3 to persist the select adn as_of, but if we don't carry it over, we'll only get the info from the left-side query, so the as_of would be there, but nto select

kbolashev: edge case with multiple selects/as_ofs, don't know how to deal with that one

dagshub/data_engine/model/query.py
kbolashev: That was how I got it to work. There are probably other ways to make it work, but this is the one I grew into organically.

Right now it's all one big Datasource class that the user is interacting with, and the DatasourceQuery is a "DTO" container for the query tree + anything that might be related to that.

The problem for me was that I didn't want to complicate things by having the interface be "multimodal", as in if you do x you get back a datasource, and if you do y, you get back a query. The complications from that would mainly be code duplication. Think this:
```python
ds.as_of(t1).all() # as_of is on the datasource
(ds["aaa"] == "bb").as_of(t1).all() # as_of in on the query
```
Same goes for `.all()` even. Basically now I'll have to duplicate every function  that you could do on a Datasource on a Query. Which is doable with a `__getattr__` magic I guess, but it feels a bit smelly

kbolashev:
Okay, after a talk, another shuffle:
- `DatasourceQuery` is now `QueryFilterTree`, only has the tree of the filters and composition logic related to that
- `DatasourceQuery` is now a higher level class that contains `as_of`, `select`, and `filter` that is a `QueryFilterTree` (but still serialized as `query` due to graphql stuff)

dagshub/data_engine/model/datasource.py
```diff
@@ -798,16 +926,19 @@ def __getitem__(self, other: Union[slice, str, "Datasource", "Field"]):
...
                 new_ds._query.compose("and", other_query)
             return new_ds
         elif type(other) is Field:
             if not self.has_field(other.field_name):
                 raise FieldNotFoundError(other.field_name)
-            new_ds._query = DatasourceQuery(other.field_name, other.as_of_timestamp)
+            new_ds._query.filter = QueryFilterTree(
+                other.field_name,
+                field_as_of=other.as_of_timestamp,
+            )
```
guysmoilov, line 941: Aren't you missing a `if self._query.filter.is_empty` check here?
If so, and it passed the tests, that means we need more tests for Field scenarios and composition!

dagshub/data_engine/model/datasource.py
```diff
@@ -1064,3 +1195,119 @@ def wrap_bytes(val: bytes) -> str:
 
     def get_metadata_entries(self):
         return self._metadata_entries
+
+
+@dataclass
+class DatasourceQuery(DataClassJsonMixin):
+    as_of: Optional[int] = field(default=None, metadata=config(exclude=exclude_if_none, letter_case=LetterCase.CAMEL))
```
guysmoilov, line 1202: Why is it an optional int if the args we accept for as_of are only datetime and float?
Shouldn't that mean we should accept ints as well?

dagshub/data_engine/model/datasource.py
```diff
@@ -1064,3 +1195,119 @@ def wrap_bytes(val: bytes) -> str:
...
+    select: Optional[List] = field(default=None, metadata=config(exclude=exclude_if_none))
```
guysmoilov, line 1203: Don't want to define the List type?

dagshub/data_engine/model/datasource.py
```diff
@@ -1064,3 +1195,119 @@ def wrap_bytes(val: bytes) -> str:
...
+    )
+
+    def __deepcopy__(self, memodict={}):
+        other = DatasourceQuery(
+            as_of=self.as_of,
+            filter=self.filter.__deepcopy__(),
+        )
+        if self.select is not None:
+            other.select = self.select.copy()
+        return other
```
guysmoilov, line 1216: Can't you use just json serder for this? Or you don't want to for efficiency?

dagshub/data_engine/model/datasource.py
```diff
@@ -1064,3 +1195,119 @@ def wrap_bytes(val: bytes) -> str:
...
+    dataset_id: Optional[Union[str, int]] = None
+    """ID of the assigned dataset"""
+    dataset_name: Optional[str] = None
+    """Name of the assigned dataset"""
+    timestamp: Optional[float] = None
+    """Timestamp of serialization"""
+    modified: Optional[bool] = None
+    """Does the query differ from the query in the assigned dataset"""
+    link: Optional[str] = None
+    """URL to open this datasource on DagsHub"""
```
guysmoilov, line 1268: It's OK that this never gets populated right now?

guysmoilov (changes requested):

dagshub/data_engine/model/datasource.py
kbolashev: The int as_of is the one that's accepted by GraphQL, the datetime/float is the user facing one.

dagshub/data_engine/model/datasource.py
kbolashev: This is running every time you add a new filter/change the datasource in any way, so yea, I try not to waste time on serializing here

dagshub/data_engine/model/datasource.py
kbolashev: Yea, I checked it, it was no big deal that it was empty

dagshub/data_engine/model/datasource.py
guysmoilov: I guess I couldn't figure out from reading the code why `DatasourceState` needs to exist apart from `DatasourceSerializedState` but I might have missed the reason.
Is there a case where you need only one and not the other, in both directions?

dagshub/data_engine/model/datasource.py
```diff
@@ -543,48 +588,118 @@ def _upload_metadata(self, metadata_entries: List[DatapointMetadataUpdateEntry])
...
+    def log_to_mlflow(self, artifact_name=DEFAULT_MLFLOW_ARTIFACT_NAME):
+        """
+        Logs the current datasource state to MLflow as an artifact.
+
+        Args:
+            artifact_name: Name of the artifact that will be stored in the MLflow run
+        """
+        run = mlflow.active_run()
+        mlflow.log_dict(self._to_dict(), artifact_name)
+        log_message(f'Saved the datasource state to MLflow (run "{run.info.run_name}") as "{artifact_name}"')
```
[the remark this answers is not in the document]
kbolashev: Done

dagshub/data_engine/model/datasource.py
```diff
@@ -543,48 +588,118 @@ def _upload_metadata(self, metadata_entries: List[DatapointMetadataUpdateEntry])
...
-            repo=self._source.repo,
-            name=self._source.name,
-            datasetQuery=self.serialize_gql_query_input()
+        path = Path(path)
+        if path.is_dir():
+            if self.assigned_dataset is not None:
+                name = self.assigned_dataset.dataset_name
+            else:
+                name = self.source.name
+            path = path / f"{name}.json"
```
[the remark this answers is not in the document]
kbolashev: Changed, sorry about it

dagshub/data_engine/model/datasource.py
kbolashev: It's a list of dicts that aren't really structured.
I'll define that it's dictionaries inside at least

dagshub/data_engine/model/datasource.py
kbolashev: Amazing catch, thank you.
Added a test that composes string filter -> field filter and vice verse

dagshub/data_engine/model/datasource.py
```diff
@@ -609,15 +610,19 @@ def save_dataset(self, name: str) -> "Datasource":
...
-        run = mlflow.active_run()
-        mlflow.log_dict(self._to_dict(), artifact_name)
+            artifact_name: Name of the artifact that will be stored in the MLflow run.
+            run: MLflow run to save to. If ``None``, uses the active MLflow run.
+        """
+        if run is None:
+            run = mlflow.active_run()
+            if run is None:
+                raise RuntimeError("No mlflow run is currently active")
+        mlflow.MlflowClient().log_dict(run.info.run_id, self._to_dict(), artifact_name)
```
guysmoilov, line 625: OK sorry for misleading you about how this should behave, but I read up now in their docs and their default behavior is different from what I expected.
1. We should be using the "fluent" API [`mlflow.log_dict`](https://github.com/mlflow/mlflow/blob/a06e82cbcc537f7f935f6198b5b180cfc59846d6/mlflow/tracking/fluent.py#L1084), which creates an active run if one doesn't exist yet. This is the behavior in all their fluent APIs like log_metric.
2. You can still get either a string or a Run object as the parameter (or None), if it's a Run object then just take its ID

OK ?

guysmoilov (changes requested):
Last small change

dagshub/data_engine/model/datasource.py
kbolashev: Oh, I don't know how I missed the `run_id` arg in that. Thank you for catching it.
I even drilled into the code to find _how_ they're logging the dict (that's where I got line 625 from), but completely glossed over the run_id in the function definition.

kbolashev: wait, second opinion: I actually looked at where I got it from, and apparently I had 2.3.2 installed, and they didn't have that arg there yet. So the run_id in the fluent API is a newer addition.
It looks like the code I have here is almost a repeat of what theirs is doing (except for the part where I don't create a run automatically). So it's your call on whether we want a version cutoff here

kbolashev:
I'm not sure if I actually like what MLflow is doing here with the "_get_or_create_run()", because I can imagine it leading newbie users accidentally creating un-closed runs, but just my 2c.

I made sure to return the run into which we saved the datasource just so there's a way to get the run to which we saved in case there was no run currently.

kbolashev:
Ok, weird.
I tried running `ds.log_to_mlflow()` without first creating a run, and the run it created is marked as `FINISHED` after the script exited.
Don't know what kind of magic they're doing, but it seems like it's not a big deal

guysmoilov (approved):

PR #416, opened 2024-01-09
Add Support ID to Data Engine errors

sdafni:
Screen Shot 2024-01-10 at 12 32 14

### DIFF
--- a/dagshub/data_engine/client/data_client.py
+++ b/dagshub/data_engine/client/data_client.py
@@ -25,6 +25,7 @@ from dagshub.data_engine.client.models import ScanOption
 from dagshub.data_engine.client.gql_mutations import GqlMutations
 from dagshub.data_engine.client.gql_queries import GqlQueries
 from dagshub.data_engine.model.datasource import Datasource, DatapointMetadataUpdateEntry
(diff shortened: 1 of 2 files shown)


kbolashev (changes requested):
Good idea with `raise ... from e` but I think it'd be preferrable if it all gets raised as one single exception.
Right now if I understand it correctly from [docs](https://docs.python.org/3/reference/simple_stmts.html#the-raise-statement), you'll basically get
```
Stack Trace (actual backend error)

The above exception was the direct cause of the following exception:

Stack Trace (the new error which is just the support ID)
```

It'd be preferrable to raise a new DataEngineGQLError that contains the original error + the support id, and the print would print both of them at the same time

kbolashev (changes requested):
Reread what I wrote about the exceptions pls.

I added a raise inside of the code so it would always trigger, here's what my stack trace looks like

image

If you encapsulate the TransportError inside of the DataEngineError (as in the constructor would be `DataEngineError(transportError, supportId)`, and the print would print both of them), you'll be able to print both the messages on the same line.
Right now you'd have to scroll way up (because the stack traces would be bigger in real life) to find the original error and what was in it. Makes it harder to solve

PR #424, opened 2024-01-30
delete metadata

sdafni:


dagshub/data_engine/model/datasource.py
```diff
@@ -543,6 +551,13 @@ def _upload_metadata(self, metadata_entries: List[DatapointMetadataUpdateEntry])
         # Update the status from dagshub, so we get back the new metadata columns
         self.source.get_from_dagshub()
 
+    def delete_metadata(self, path, id, key):
+        metadata_entries = [DatapointDeleteMetadataEntry(datapointId=id, url=path, key=key)]
+        self.source.client.delete_metadata(self, metadata_entries)
+
+        # Update the status from dagshub, so we get back the new metadata columns
+        self.source.get_from_dagshub()
+
```
kbolashev, line 560: Considering that we also want to be able to delete the whole field from the datasource at once, this needs to be named differently at the very least.

Delete field from the whole datasource: `delete_metadata`
Delete field from just one datapoint: `delete_metadata_from_datapoint`

But honestly I think it's fine if this function exists just on the datapoint object, even if that kind of forces users to do a query before deletion.

Another problem with this function: `id` is not user-usable, because it requires a user to go through a query anyway. If you _do_ want to keep this function, then either the GQL mutation shouldn't accept BOTH path and id, but either of them instead, 
Alternative that doesn't require redesigning the GQL mutation [less preferrable IMO]: do a roundtrip here with `(ds[path] == "<url>").head()[0]`. (I don't think we still have this as a "neat" function like `get_datapoint()` that you could use). Also keep in mind that `self` also carries with it a filter, so doing a query like that would require a copy of the `self` with its filter reset
Alternative alternative: make this function accept `Datapoint` instead of `ID+Path`

dagshub/data_engine/model/datasource.py
```diff
@@ -543,6 +551,13 @@ def _upload_metadata(self, metadata_entries: List[DatapointMetadataUpdateEntry])
         # Update the status from dagshub, so we get back the new metadata columns
         self.source.get_from_dagshub()
 
+    def delete_metadata(self, path, id, key):
+        metadata_entries = [DatapointDeleteMetadataEntry(datapointId=id, url=path, key=key)]
+        self.source.client.delete_metadata(self, metadata_entries)
+
+        # Update the status from dagshub, so we get back the new metadata columns
+        self.source.get_from_dagshub()
```
kbolashev, line 559: I don't see anything from frontend/backend that would make it so the whole metadata field is disappearing, unless I'm missing something.
If the metadata field can't disappear from the whole datasource after deletion, then there's no need to `get_from_dagshub()`

kbolashev (changes requested):
Context for the big comment: Consider that I as a user can do after this PR:

```python
ds = datasources.get_datasource(...)
ds.delete_metadata(???, ???, "field")
```

How would you expect that to work?

dagshub/data_engine/model/datasource.py
sdafni: i ment this to not be a user function , i should rename it _delete_metadata.
the api is datapoint.delete_metadata(name)

kbolashev:
Change the `_of_datapoint` to `_from_datapoint` pls

dagshub/data_engine/model/datapoint.py
```diff
@@ -45,8 +45,15 @@ def __getitem__(self, item):
...
 
-    def delete_metadata(self, name):
-        self.datasource.delete_metadata(self.path, self.datapoint_id, name)
+    def delete_metadata_field(self, name):
+        """
+        delete a metadata field from this datapoint
+        the deleted value can be accessed using versioned query
+        with time before the deletion
+        Args:
+             name - name of field
```
kbolashev, line 54: Change arg name from name -> field

```suggestion
        Delete a metadata field from this datapoint.
        The deleted value can be accessed using versioned query
        with time set before the deletion.
        
        Args:
             field - field to delete
```

dagshub/data_engine/model/datasource.py
```diff
@@ -597,14 +595,23 @@ def _upload_metadata(self, metadata_entries: List[DatapointMetadataUpdateEntry])
...
-        metadata_entries = [DatapointDeleteMetadataEntry(datapointId=id, url=path, key=key)]
-        self.source.client.delete_metadata(self, metadata_entries)
+    def delete_metadata_from_datapoints(self, datapoints: List[Datapoint], names: List[str]):
+        """
+        delete a metadata fields from these datapoints
+        the deleted values can be accessed using versioned query
+        with time before the deletion
+        Args:
+             datapoints - data points to act upon
+             names - name of fields to delete
```
kbolashev, line 605: Same comments as on the datapoint. Make sure the docs looks nice, and run the sphinx build

kbolashev (changes requested):

kbolashev (approved):

PR #425, opened 2024-01-31
delete datapoint

sdafni:


dagshub/data_engine/client/data_client.py
```diff
@@ -205,11 +206,31 @@ def update_metadata(self, datasource: Datasource, entries: List[DatapointMetadat
...
+        """
+        Delete a metadata from a datapoint
+
+        Args:
+            datasource (Datasource): The datasource instance to be updated
+            entry: the datapoint to delete
+
+        Returns:
+            Updates the Datasource.
+
```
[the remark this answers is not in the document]
kbolashev: `Returns:` should be explaining what is the object that is being returned.
"Updates the Datasource" is a side-effect of executing the action, and it's not even _actually_ doing anything with the `Datasource` object

dagshub/data_engine/model/datapoint.py
```diff
@@ -45,11 +45,34 @@ def __getitem__(self, item):
...
+        prompt = (
+            f'You are about to delete the datapoint [{self.path}]."\n'
+            f"This will remove the datapoint and metadata from unversioned queries, "
+            f"but won't delete the underlying file."
+        )
+        if not force:
+            user_response = prompt_user(prompt)
+            if not user_response:
+                print("Deletion cancelled")
+                return
```
kbolashev, line 73: Skip this and pass the force arg to `delete_datapoints` instead

dagshub/data_engine/model/datasource.py
```diff
@@ -595,10 +595,38 @@ def _upload_metadata(self, metadata_entries: List[DatapointMetadataUpdateEntry])
...
+        - Datasource scanning will *not* add these datapoints back.
+
+        Args:
+            datapoints: list of datapoints objects to delete
+            force: Skip the confirmation prompt
+        """
+        prompt = (
+            f'You are about to delete the datapoints {[d.path for d in datapoints]}."\n'
+            f"This will remove the datapoint and metadata from unversioned queries, "
+            f"but won't delete the underlying file."
```
kbolashev, line 618: Print out the paths on a separate line with a \t added.
Something like (didn't test it, just typing it out here)
```suggestion
            f"You are about to delete the following datapoint(s):\n"
            f"{'\n'.join(['\t'+d.path for d in datapoints])}\n"
            f"This will remove the datapoint(s) and metadata from unversioned queries, "
            f"but won't delete the underlying file(s)."
```

kbolashev (changes requested):

PR #430, opened 2024-02-13
dagshub upload: variable batch size.

### LINKED ISSUE #429: upload failing on dataset with large files
My dataset - ~640GB - is failing on dagshub upload.
This is apparently due to some extremely large files (~70-80 MB) being grouped together to a single upload step of ~7GB.

yairl:
1. Uploads are now grouped into the minimum of 100 files or 100MB-s each (not configurable).
2. All uploaded files are discovered before upload begins, to make the global progress bar more useful.
3. Minor restructuring to make the code cleaner.

Notes:

- On my machine (Debian 12), the progress bar is printed many times and not just updated in place. This is annoying, wondering whether you'd like that changed as well.
- In process prints such as "Upload finished successfully!" are confusing and polluting the screen when it's one part of a bigger upload.
- With this change, you may want to change the file batch size from 100 to a larger number, or remove completely.

Resolves #429.

### DIFF
--- a/dagshub/upload/wrapper.py
+++ b/dagshub/upload/wrapper.py
@@ -296,6 +296,7 @@ class Repo:
         new_branch: str = None,
         last_commit: str = None,
         force: bool = False,
+        quiet: bool = False,
     ):
         """
         Upload a list of binary files to the specified directory.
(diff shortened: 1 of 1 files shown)


yairl:
@kbolashev sorry for the "batchs" mistake.
I hate these :)

kbolashev (approved):
Thanks a lot for doing this, this is absolutely better now!!!

There's probably still things that can be done, for example not bounding batches by folders, but I honestly don't remember if our uploading API supports that.

I took a liberty of changing a couple variable names to be more descriptive, hope that's ok.

I also liked the suggestion about not cluttering the output that much on dir upload, so I hide those now when a directory is being uploaded. Pull the changes and see if you like it :)

PR #432, opened 2024-02-15
Feature/Support sorting

simonlsk:
* Add order_by capability
* Add introspection to check capability available.

### DIFF
--- a/dagshub/data_engine/client/data_client.py
+++ b/dagshub/data_engine/client/data_client.py
@@ -12,6 +12,7 @@ import dagshub.common.config
 from dagshub.common import config
 from dagshub.common.analytics import send_analytics_event
 from dagshub.common.rich_util import get_rich_progress
+from dagshub.data_engine.client.gql_introspections import GqlIntrospections
 from dagshub.data_engine.client.models import (
     DatasourceResult,
     DatasourceType,
@@ -189,21 +190,7 @@ class DataClient:
 
     @cached_property
     def introspection_dict(self):
-        introspection = """
-{
-  __schema {
-    types {
-      name
-      inputFields {
-        name
-        type {
-          name
-        }
-      }
-    }
-  }
-}
-        """
+        introspection = GqlIntrospections.input_fields()
         return self._exec(introspection)
 
     def update_metadata(self, datasource: "Datasource", entries: List["DatapointMetadataUpdateEntry"]):
--- a/dagshub/data_engine/client/data_client.py
+++ b/dagshub/data_engine/client/data_client.py
@@ -238,7 +238,7 @@ class DataClient:
         Returns:
             List[DatasourceResult]: A list of datasources that match the filtering criteria.
         """
-        q = GqlQueries.datasource()
+        q = GqlQueries.datasource().generate()
         params = GqlQueries.datasource_params(id=id, name=name)
 
         res = self._exec(q, params)["datasource"]
@@ -305,7 +305,7 @@ class DataClient:
         Returns:
             List[DatasetResult]: A list of datasets that match the filtering criteria.
         """
-        q = GqlQueries.dataset()
+        q = GqlQueries.dataset().generate()
         params = GqlQueries.dataset_params(id=id, name=name)
 
         res = self._exec(q, params)["dataset"]
--- a/dagshub/data_engine/client/data_client.py
+++ b/dagshub/data_engine/client/data_client.py
@@ -12,7 +12,7 @@ import dagshub.common.config
 from dagshub.common import config
 from dagshub.common.analytics import send_analytics_event
 from dagshub.common.rich_util import get_rich_progress
-from dagshub.data_engine.client.gql_introspections import GqlIntrospections
+from dagshub.data_engine.client.gql_introspections import GqlIntrospections, QueryInputIntrospection
 from dagshub.data_engine.client.models import (
     DatasourceResult,
     DatasourceType,
@@ -185,13 +185,14 @@ class DataClient:
             first=limit,
             after=after,
         )
-        q.validate_params(params, self.introspection_dict)
+        q.validate_params(params, self.query_introspection)
         return self._exec(q.generate(), params)["datasourceQuery"]
 
     @cached_property
-    def introspection_dict(self):
+    def query_introspection(self) -> QueryInputIntrospection:
         introspection = GqlIntrospections.input_fields()
-        return self._exec(introspection)
+        introspection_dict = self._exec(introspection)
+        return dacite.from_dict(data_class=QueryInputIntrospection, data=introspection_dict["__schema"])
 
     def update_metadata(self, datasource: "Datasource", entries: List["DatapointMetadataUpdateEntry"]):
         """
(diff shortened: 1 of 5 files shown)


dagshub/data_engine/model/datasource.py
```diff
@@ -334,6 +334,38 @@ def as_of(self, time: Union[float, datetime.datetime]) -> "Datasource":
         new_ds._query.as_of = to_timestamp(time)
         return new_ds
 
+    def order_by(self, *args: Union[str, Tuple[str, bool], Tuple[str, str]]):
```
kbolashev, line 337:
```suggestion
    def order_by(self, *args: Union[str, Tuple[str, Union[bool, str]]]) -> "Datasource":
```

dagshub/data_engine/model/datasource.py
```diff
@@ -334,6 +334,38 @@ def as_of(self, time: Union[float, datetime.datetime]) -> "Datasource":
...
+            ds.sort_by("size", ("date", False)).all()
+        """
+        new_ds = self.__deepcopy__()
+        new_ds.get_query().order_by = []
+        for arg in args:
+            if isinstance(arg, str):
+                new_ds.get_query().order_by.append({"field": arg, "order": "ASC"})
+            else:
+                if len(arg) != 2:
+                    raise RuntimeError(f"Invalid sort argument {arg}")
```
kbolashev, line 359:
```suggestion
                if len(arg) != 2:
                    raise RuntimeError(f"Invalid sort argument {arg}. Needs to be a tuple of (<field>, 'asc'|'desc'|True|False)")
```

dagshub/data_engine/model/datasource.py
```diff
@@ -334,6 +334,38 @@ def as_of(self, time: Union[float, datetime.datetime]) -> "Datasource":
...
+                if isinstance(arg[1], bool):
+                    order = "ASC" if arg[1] else "DESC"
+                elif isinstance(arg[1], str) and arg[1].upper() in ["ASC", "DESC"]:
+                    order = arg[1].upper()
+                else:
+                    raise RuntimeError(f"Invalid sort argument {arg}")
```
kbolashev, line 365:
```suggestion
                else:
                    raise RuntimeError(f"Invalid sort argument {arg}. Second value needs to be either of 'asc'|'desc'|True|False")
```

dagshub/data_engine/model/datasource.py
```diff
@@ -334,6 +334,38 @@ def as_of(self, time: Union[float, datetime.datetime]) -> "Datasource":
...
+        Args:
+            Fields to sort by. Can be either of:
+                - Name of the field to sort by: ``"field"``.
+                - A tuple of ``(field_name, ascending)``: ``("field", True)``.
+                - A tuple of ``(field_name, "asc"|"desc")``: ``("field", "asc")``.
+
+        Examples::
+
+            ds.sort_by("size", ("date", "desc")).all()
+            ds.sort_by("size", ("date", False)).all()
```
kbolashev, line 350:
```suggestion
            ds.order_by("size", ("date", "desc")).all()
            ds.order_by("size", ("date", False)).all()
```

dagshub/data_engine/model/datasource.py
```diff
@@ -334,6 +334,38 @@ def as_of(self, time: Union[float, datetime.datetime]) -> "Datasource":
...
+        """
+        new_ds = self.__deepcopy__()
+        new_ds.get_query().order_by = []
```
kbolashev, line 353: Conflicted on whether clearing is a good idea. I imagine a user that doesn't know you could order_by multiple columns might do
```python
ds.order_by("size").order_by("is_val")
```

On the other hand if you don't reset it here, then there's no way to reset a sort 🤔 

You should 100% say that calling the function resets whatever sort was set previously
By the way, you should also say that doing an `ds.order_by()` resets the sort to the default (datapoint addition order)

dagshub/data_engine/model/datasource.py
```diff
@@ -334,6 +334,38 @@ def as_of(self, time: Union[float, datetime.datetime]) -> "Datasource":
...
+            else:
+                if len(arg) != 2:
+                    raise RuntimeError(f"Invalid sort argument {arg}")
+                if isinstance(arg[1], bool):
+                    order = "ASC" if arg[1] else "DESC"
+                elif isinstance(arg[1], str) and arg[1].upper() in ["ASC", "DESC"]:
+                    order = arg[1].upper()
+                else:
+                    raise RuntimeError(f"Invalid sort argument {arg}")
+                new_ds.get_query().order_by.append({"field": arg[0], "order": order})
```
kbolashev, line 366: Extending as you go might lead to broken ordering later (even though I see you clearing it here right now).
Maybe have an orders list and then set `new_ds.get_query().order_by = orders` in the end

dagshub/data_engine/client/data_client.py
```diff
@@ -180,8 +184,27 @@ def _datasource_query(
...
+        name
+        type {
+          name
+        }
+      }
+    }
+  }
+}
+        """
+        return self._exec(introspection)
```
kbolashev, line 207: I know that this is a "do it right now" thing, but I want it to be done okay-ishly so we're not stuck with filtering through dict.

Make the result of this into a Dacite model. That way we can have the validators on that model object.

dagshub/data_engine/client/gql_queries.py
```diff
@@ -41,7 +41,7 @@ def datasource_params(id: Optional[Union[int, str]], name: Optional[str]) -> Dic
 
     @staticmethod
     @functools.lru_cache()
-    def datasource_query(include_metadata: bool) -> str:
+    def datasource_query(include_metadata: bool) -> GqlQuery:
```
kbolashev, line 44: If you're changing the return type on this query, change it on every other query/mutation too.

kbolashev (changes requested):
I want to play a bit with the introspection myself.
This looks good already, but I don't want us to be trudging through dicts for input validation

dagshub/data_engine/model/datasource.py
kbolashev: I would also probably make them a bit different from each other
```
            ds.order_by("size").all()                              # Order by ascending size
            ds.order_by("date", "desc"), "size").all()   # Order by descending date, then ascending size
```

dagshub/data_engine/model/datasource.py
kbolashev: I guess since raising an error will not return `new_ds` it actually doesn't matter. nevermind

dagshub/data_engine/client/gql_queries.py
```diff
@@ -70,8 +70,31 @@ def datasource_query(include_metadata: bool) -> str:
...
+
+            # Get sent fields
+            query_input = params.get("queryInput")
+            if query_input is None:
+                return
+            sent_fields = query_input.keys()
+            # Check serialized query input fields exist in introspection
+            if not all([f in introspect_query_input_fields for f in sent_fields]):
+                unsupported_fields = [f for f in sent_fields if f not in introspect_query_input_fields]
+                raise ValueError(f"QueryInput fields are not supported: {unsupported_fields}")
```
kbolashev, line 96: This shouldn't be an inner function, fine enough to take it outside imo

kbolashev (changes requested):

dagshub/data_engine/model/datasource.py
simonlsk: I actually like your suggestion.

kbolashev (approved):

PR #440, opened 2024-02-26
feature/support select fields

simonlsk:
# Spec
* Add `selectFields` field in the `datasourceQuery`.
* First introspect if field is available and require it only if possible.


dagshub/data_engine/client/gql_introspections.py
```diff
@@ -7,55 +7,70 @@
...
     name: str
-    inputFields: List[InputField]
+    inputFields: List[Field]
+    fields: List[Field]
+
+    def __hash__(self):
+        return hash(self.hash_str())
+
+    def hash_str(self):
+        return self.name + "".join([f.name for f in self.fields]) + "".join([f.name for f in self.inputFields])
```
kbolashev, line 59: If you're doing all the hash stuff for the contain check at line 104, then it won't really work, you need to put them in a set and not in a list for that. List containment doesn't utilize hashing iirc.

kbolashev: Unless I'm missing something and there's another reason you're doing the hashing. Please point me at that if it is so

kbolashev (approved):
So this isn't really being exposed to the user from what I read, right?

dagshub/data_engine/client/gql_introspections.py
simonlsk: It’s necessary for functools lru_cache to work since it’s passed as an argument to a cached function.

PR #442, opened 2024-03-02
Add s3 upload functionality in `dagshub.upload` and in cli `dagshub upload`

deanp70:
Needed this for the weekend project so I implemented it myself.
Support parallel upload with boto3 for files and directories while preserving folder structure.

Supported in both python and cli upload functions.

I should probably add a test, @kbolashev would appreciate your guidance on that.

### DIFF
--- a/dagshub/common/cli.py
+++ b/dagshub/common/cli.py
@@ -201,8 +201,11 @@ def download(
 @click.option(
     "--versioning", help="Versioning system to be used to upload the file(s)", type=click.Choice(["git", "dvc", "auto"])
 )
+@click.option(
+    "--to-bucket", is_flag=True, help="Upload the file(s) to the repo's DagsHub Storage bucket (s3-compatible)"
+)
 @click.pass_context
-def upload(ctx, filename, target, repo, message, branch, verbose, update, quiet, host, versioning, **kwargs):
+def upload(ctx, filename, target, repo, message, branch, verbose, update, quiet, host, versioning, to_bucket, **kwargs):
     """
     Upload FILENAME to REPO at location TARGET.
 
@@ -224,7 +227,12 @@ def upload(ctx, filename, target, repo, message, branch, verbose, update, quiet,
     repo = Repo(owner=owner, name=repo_name, branch=branch)
     try:
         repo.upload(
-            local_path=filename, remote_path=target, commit_message=message, force=update, versioning=versioning
+            local_path=filename,
+            remote_path=target,
+            commit_message=message,
+            to_bucket=to_bucket,
+            force=update,
+            versioning=versioning,
         )
     except UpdateNotAllowedError:
         log_message(
--- a/dagshub/common/cli.py
+++ b/dagshub/common/cli.py
@@ -202,10 +202,10 @@ def download(
     "--versioning", help="Versioning system to be used to upload the file(s)", type=click.Choice(["git", "dvc", "auto"])
 )
 @click.option(
-    "--to-bucket", is_flag=True, help="Upload the file(s) to the repo's DagsHub Storage bucket (s3-compatible)"
+    "--bucket", is_flag=True, help="Upload the file(s) to the repo's DagsHub Storage bucket (s3-compatible)"
 )
 @click.pass_context
-def upload(ctx, filename, target, repo, message, branch, verbose, update, quiet, host, versioning, to_bucket, **kwargs):
+def upload(ctx, filename, target, repo, message, branch, verbose, update, quiet, host, versioning, bucket, **kwargs):
     """
     Upload FILENAME to REPO at location TARGET.
 
@@ -230,7 +230,7 @@ def upload(ctx, filename, target, repo, message, branch, verbose, update, quiet,
             local_path=filename,
             remote_path=target,
             commit_message=message,
-            to_bucket=to_bucket,
+            bucket=bucket,
             force=update,
             versioning=versioning,
         )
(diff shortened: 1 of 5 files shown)


dagshub/common/cli.py
```diff
@@ -201,8 +201,11 @@ def download(
 @click.option(
     "--versioning", help="Versioning system to be used to upload the file(s)", type=click.Choice(["git", "dvc", "auto"])
 )
+@click.option(
+    "--to-bucket", is_flag=True, help="Upload the file(s) to the repo's DagsHub Storage bucket (s3-compatible)"
+)
```
kbolashev, line 206:
```suggestion
@click.option(
    "--bucket", is_flag=True, help="Upload the file(s) to the repo's DagsHub Storage bucket (s3-compatible)"
)
```

dagshub/common/cli.py
```diff
@@ -201,8 +201,11 @@ def download(
...
 @click.pass_context
-def upload(ctx, filename, target, repo, message, branch, verbose, update, quiet, host, versioning, **kwargs):
+def upload(ctx, filename, target, repo, message, branch, verbose, update, quiet, host, versioning, to_bucket, **kwargs):
```
kbolashev, line 208:
```suggestion
def upload(ctx, filename, target, repo, message, branch, verbose, update, quiet, host, versioning, bucket, **kwargs):
```

dagshub/common/cli.py
kbolashev: Change the `to-bucket` to `bucket` everywhere basically

dagshub/upload/wrapper.py
```diff
@@ -549,6 +561,66 @@ def get_repo_url(self, url_format, directory, branch=None):
...
+        """
+        Upload a single file to S3.
+
+        :param s3_client: An instance of boto3 S3 client
+        :param local_path: The local path to the file
+        :param bucket_name: The S3 bucket name
+        :param remote_path: The S3 path where the file will be uploaded
+        """
+        s3_client.upload_file(local_path, bucket_name, remote_path)
+        log_message(f'Uploaded {os.path.relpath(local_path, os.getcwd())} to "{self._api.full_name}"...', logger)
```
kbolashev, line 574: Don't spam the logs for every file pls

dagshub/upload/wrapper.py
```diff
@@ -549,6 +561,66 @@ def get_repo_url(self, url_format, directory, branch=None):
...
+        """
+        Upload a file or directory to an S3 bucket, preserving the directory structure.
+
+        :param local_path: Path to the local directory to upload
+        :param remote_path: The directory path within the S3 bucket
+        :param recursive: True if this is a directory upload, or False if single file
+        :param max_workers: The maximum number of threads to use
+        """
+        file_exceptions = ['.DS_Store']  # Default list of files to skip
+
```
kbolashev, line 593: I understand the frustration with `.DS_Store`, but we shouldn't be enforcing that onto users imo.
This could be hidden behind a `respect_gitignore` flag maybe.

BTW, this reminds me, that `.dvcignore` exists and we don't respect that for the uploading either 🙃

dagshub/upload/wrapper.py
```diff
@@ -549,6 +561,66 @@ def get_repo_url(self, url_format, directory, branch=None):
...
+        s3 = get_repo_bucket_client(self._api.full_name)
+        if remote_path.split("/")[0] == self.name:
+            remote_path = remote_path.split("/", 1)[1]
+        if not recursive:
+            self.upload_file_to_s3(s3_client=s3, local_path=local_path, bucket_name=self.name, remote_path=remote_path)
```
kbolashev, line 598: This isn't what `recursive` usually means.

The logic should be:
```python
if is_file(local_path):
  upload_file(file)
if is_dir(local_path):
  if recursive:
    <the os.walk you have written>
  else:
    for file in os.listdir(local_path):
      upload_file(file)
```

I recommend looking at the `download_files` function for a way to make it more manageable, basically splitting the process into two steps: 1) get the list of things to upload and where to upload them, 2) actually upload them

dagshub/upload/wrapper.py
```diff
@@ -549,6 +561,66 @@ def get_repo_url(self, url_format, directory, branch=None):
...
+            with ThreadPoolExecutor(max_workers=max_workers) as executor:
+                futures = [executor.submit(
+                    self.upload_file_to_s3,
+                    s3_client=s3,
+                    local_path=task[0],
+                    bucket_name=self.name,
+                    remote_path=task[1],
+                ) for task in upload_tasks]
+                for future in futures:
+                    future.result()  # Wait for all futures to complete
```
kbolashev, line 622: Try to copy what I did in `download_files` with the progress bar.
Honestly it turned out pretty complicated, so I can try to transplant it myself if needed

setup.py
```diff
@@ -41,7 +41,8 @@ def get_version(rel_path: str) -> str:
     "treelib~=1.6.4",
     "pathvalidate~=3.0.0",
     "python-dateutil",
-    "tenacity~=8.2.3"
+    "tenacity~=8.2.3",
+    "boto3~=1.34.54"
```
kbolashev, line 45: Still not sure if we should pin the versions or not, considering we want to play nice with other libraries that get installed, and A LOT of libraries install boto. Might break pip install for dependencies.
The solution I used before that was to hide the boto behind a lazy import, so the user gets an exception to `pip install boto`. Used that for the bucket clients and for the fiftyone imports.
However this feature will probably be used more often than those... I'll mull over what exactly to do here

kbolashev (changes requested):
Thank you! Looks really neat and useful

setup.py
kbolashev: Reading up on this: https://discuss.python.org/t/should-i-be-pinning-my-dependencies/13159

kbolashev (approved):

PR #444, opened 2024-03-07
Add ability to visualize in the dagshub gallery view

kbolashev:
Adds a parameter `visualizer` param to the `visualize` function, that either shows the thing in dagshub or fiftyone. Default is still fiftyone for now to keep backwards compat.
 <BREAKING> once major version is bumped, set the default of the visualize to "dagshub"

### DIFF
--- a/dagshub/common/__init__.py
+++ b/dagshub/common/__init__.py
@@ -1,21 +1,5 @@
 import rich.console
-
-
-def is_inside_notebook():
-    try:
-        from IPython import get_ipython
-
-        return get_ipython() is not None
-    except ModuleNotFoundError:
-        return False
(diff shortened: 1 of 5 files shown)


kbolashev:
[Python Client Data Engine: allow visualize() into dagshub UI](https://trello.com/c/b6MQEaOh/1973-python-client-data-engine-allow-visualize-into-dagshub-ui)

simonlsk (approved):
LGTM,
Default seems to be DagsHub here or did I miss something?

kbolashev:
Oops, forgot that I reset the default too... Should probably set it to fiftyone and then flip it later

PR #447, opened 2024-03-11
Loading a model from DagsHub repo using `get_model_path`

kbolashev:
This pull request adds a function to streamline loading models from a DagsHub Repository: `dagshub.models.get_model_path`.

This function tries to find a model in the repository and loads it to a local directory, allowing the user to use it with e.g. transformers:

```python
AutoModel.load_pretrained(get_model_path('user/repo'))
```

Order of lookup for the model dir:
- `.dagshub/model.yaml` file with a `model_dir` key
- `model` and `models` dirs in repo root
- `model` and `models` dirs in DagsHub Storage root

Also has `path` and `bucket` arguments that allows the user to load from a specific path/bucket.

The loading has two download modes: `lazy` and `eager`. `lazy` runs install_hooks() in the dir with the model, and `eager` downloads files into the dir. The resulting paths are consistent between them.

The default location for saving is `~/dagshub/models/<user>/<repo>/`, and every model is always stored with its full path in the repository.

Unsolved problems:
- Model invalidation. Right now, I'm not deleting the downloaded files, so once the model is loaded to a destination, it is there for good and it won't be replaced, until the user manually deletes it.

### DIFF
--- a/dagshub/models/model_loaders.py
+++ b/dagshub/models/model_loaders.py
@@ -22,7 +22,6 @@ class ModelLoader:
     def _eager_load(self, download_dest: Path) -> Path:
         ...
 
-    @abstractmethod
     def _lazy_load(self, download_dest: Path) -> Path:
         fs = DagsHubFilesystem(project_root=download_dest, repo_url=self.repo_api.repo_url)
         res_path = download_dest / self.model_path
--- a/dagshub/models/model_loaders.py
+++ b/dagshub/models/model_loaders.py
@@ -46,7 +46,10 @@ class RepoModelLoader(ModelLoader):
         self.path = path
 
     def _eager_load(self, download_dest: Path) -> Path:
-        raise NotImplementedError
+        remote_path = self.path
+        local_path = download_dest / self.model_path
+        self.repo_api.download(remote_path, local_path=local_path, redownload=True)
+        return local_path
 
     @property
     def model_path(self) -> Path:
@@ -59,7 +62,10 @@ class DagsHubStorageModelLoader(ModelLoader):
         self.path = path
 
     def _eager_load(self, download_dest: Path) -> Path:
-        raise NotImplementedError
+        remote_path = PurePosixPath(f"s3:/{self.repo_api.repo_name}") / self.path
+        local_path = download_dest / self.model_path
+        self.repo_api.download(remote_path, local_path=local_path, redownload=True)
+        return local_path
 
     @property
     def model_path(self) -> Path:
@@ -72,7 +78,11 @@ class BucketModelLoader(ModelLoader):
         self.path = path
 
     def _eager_load(self, download_dest: Path) -> Path:
-        raise NotImplementedError
+        # Need to change from s3/bucket/bla-bla to s3:/bucket/bla-bla for download to work
+        remote_path = str(self.path).replace("/", ":/", 1)
+        local_path = download_dest / self.model_path
+        self.repo_api.download(remote_path, local_path=local_path, redownload=True)
+        return local_path
 
     @property
     def model_path(self) -> Path:
--- a/dagshub/models/model_loaders.py
+++ b/dagshub/models/model_loaders.py
@@ -11,7 +11,12 @@ class ModelLoader:
     def __init__(self, repo_api: RepoAPI):
         self.repo_api = repo_api
 
-    def load_model(self, mode: Literal["eager", "lazy"], download_dest: Path, revision: Optional[str] = None) -> Path:
+    def load_model(
+        self,
+        mode: Literal["eager", "lazy"],
+        download_dest: Path,
+        revision: Optional[str] = None,
+    ) -> Path:
         if mode == "eager":
             return self._eager_load(download_dest)
         elif mode == "lazy":
@@ -21,15 +26,14 @@ class ModelLoader:
         raise RuntimeError(f"Unknown model load mode [{mode}]")
 
     @abstractmethod
-    def _eager_load(self, download_dest: Path) -> Path:
-        ...
+    def _eager_load(self, download_dest: Path) -> Path: ...
 
     def _lazy_load(self, download_dest: Path, revision: str) -> Path:
         fs = DagsHubFilesystem(
             project_root=download_dest,
             repo_url=self.repo_api.repo_url,
             branch=revision,
-            frameworks=["transformers"]
+            frameworks=["transformers"],
         )
         res_path = download_dest / self.model_path
         log_message(
@@ -41,8 +45,7 @@ class ModelLoader:
 
     @property
     @abstractmethod
-    def model_path(self) -> Path:
-        ...
+    def model_path(self) -> Path: ...
 
 
 class RepoModelLoader(ModelLoader):
@@ -51,10 +54,23 @@ class RepoModelLoader(ModelLoader):
         self.revision = revision
         self.path = path
 
+    def load_model(
+        self,
+        mode: Literal["eager", "lazy"],
+        download_dest: Path,
+        revision: Optional[str] = None,
+    ) -> Path:
+        """Overrides the original to change the revision used. If it's not set, use the revision from the class"""
+        if revision is None:
+            revision = self.revision
+        return super().load_model(mode, download_dest, revision)
+
     def _eager_load(self, download_dest: Path) -> Path:
         remote_path = self.path
         local_path = download_dest / self.model_path
-        self.repo_api.download(remote_path, local_path=local_path)
+        self.repo_api.download(
+            remote_path, local_path=local_path, revision=self.revision
+        )
         return local_path
 
     @property
--- a/dagshub/models/model_loaders.py
+++ b/dagshub/models/model_loaders.py
@@ -78,22 +78,6 @@ class RepoModelLoader(ModelLoader):
         return Path(self.path)
 
 
-class DagsHubStorageModelLoader(ModelLoader):
-    def __init__(self, repo_api: RepoAPI, path: PurePosixPath):
-        super().__init__(repo_api)
-        self.path = path
-
-    def _eager_load(self, download_dest: Path) -> Path:
-        remote_path = PurePosixPath(f"s3:/{self.repo_api.repo_name}") / self.path
-        local_path = download_dest / self.model_path
-        self.repo_api.download(remote_path, local_path=local_path)
-        return local_path
-
-    @property
-    def model_path(self) -> Path:
-        return Path(".dagshub") / "storage" / "s3" / self.repo_api.repo_name / self.path
-
-
 class BucketModelLoader(ModelLoader):
     def __init__(self, repo_api: RepoAPI, path: PurePosixPath):
         super().__init__(repo_api)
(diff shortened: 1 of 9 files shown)


kbolashev:
WDYT: changing the name from `get_model_path` to `get_model`

dagshub/models/model_locator.py
```diff
@@ -0,0 +1,320 @@
...
+
+# TODO:
+"""
+Ways to get model:
+- automatic - with the cascade
+- from path
+- from bucket
+- from mlflow model
+- from mlflow artifact
+"""
```
simonlsk, line 29: What is left to do? 
Is it part of the PR?

dagshub/models/model_locator.py
```diff
@@ -0,0 +1,320 @@
...
+
+# TODO: don't forget to separate dvc downloads by the hash of the dir
```
simonlsk, line 31: Is this intention for future or for this PR?

dagshub/models/model_locator.py
```diff
@@ -0,0 +1,320 @@
...
+
+import dacite
+import yaml
+from pathvalidate import sanitize_filepath
+
+from dagshub.common.analytics import send_analytics_event
+from dagshub.common.api import RepoAPI
+from dagshub.common.api.repo import PathNotFoundError
+from dagshub.common.api.responses import StorageAPIEntry
+from dagshub.common.determine_repo import determine_repo
```
simonlsk, line 17: General preference, not CTA right now - Would be nice to export the function at the `common` level then use it in the code as `dagshub.common.determine_repo`.

dagshub/models/model_locator.py
```diff
@@ -0,0 +1,320 @@
...
+        self.mlflow_artifact = mlflow_artifact
+        self._download_dest = download_dest
+        self.download_type = download_type
+
+    @cached_property
+    def repo_api(self) -> RepoAPI:
+        if isinstance(self._repo, RepoAPI):
+            return self._repo
+        if self._repo is None:
+            repo_api, git_ref = determine_repo()
```
simonlsk, line 78: What if repo is passed as string what happens?

dagshub/models/model_locator.py
```diff
@@ -0,0 +1,320 @@
...
+    def git_ref(self):
+        if self._git_ref is None:
+            self._git_ref = self.repo_api.default_branch
+        return self._git_ref
+
+    @cached_property
+    def download_destination(self) -> Path:
+        if self._download_dest is not None:
+            return Path(self._download_dest)
+        return Path(sanitize_filepath(os.path.join(Path.home(), "dagshub", "models", self.repo_api.full_name)))
```
simonlsk, line 94: Should this be customizable with an env variable? 
Is there already a similar path for dagshub resources? Should this use a common logic with the token cache location for example?

dagshub/models/model_locator.py
```diff
@@ -0,0 +1,320 @@
...
+
+        if "/" not in str_path:
+            return path, StorageType.Repo
+
+        if str_path.startswith("dagshub_storage/"):
+            return str_path[len("dagshub_storage/"):], StorageType.DagshubStorage
+        for storage in self.repo_storages:
+            if str_path.startswith(f"{storage.name}/"):
+                bucketPath = f"{storage.protocol}/{str_path}"
+                return bucketPath, StorageType.Bucket
```
simonlsk, line 119: User here needs some advanced knowledge of necessary naming convention.

dagshub/models/model_locator.py
```diff
@@ -0,0 +1,320 @@
...
+        for storage in self.repo_storages:
+            if storage.name.startswith(bucket):
+                bucketPath = PurePosixPath(f"{storage.protocol}/{storage.name}")
+                break
+
+        if bucketPath is None:
+            return None
+
+        for lookup_dir in self.lookup_dirs:
+            modelPath = bucketPath / lookup_dir
```
simonlsk, line 180: The model path always comes as a directory?

dagshub/models/model_locator.py
```diff
@@ -0,0 +1,320 @@
...
+        return None
+
+    def find_model(self) -> ModelLoader:
+        model_loader: Optional[ModelLoader]
+
+        def return_if_exists_throw_otherwise():
+            if model_loader is None:
+                raise ModelNotFoundError
+            if model_loader is not None:
+                return model_loader
```
simonlsk, line 209: I don't really see the point in that function you only use it in one return point, with a simple if-else.

tests/model_loading/test_download.py
simonlsk, line 1: Needed for the tests?

simonlsk (approved):
LGTM, 
Very limited remarks and potential bugs.

dagshub/models/model_loaders.py
```diff
@@ -0,0 +1,104 @@
...
+            repo_url=self.repo_api.repo_url,
+            branch=revision,
+            frameworks=["transformers"]
+        )
+        res_path = download_dest / self.model_path
+        log_message(
+            f"Running install_hooks() in dir {download_dest}, "
+            f"run dagshub.streaming.uninstall_hooks() if you want to run another model later"
+        )
+        fs.install_hooks()
```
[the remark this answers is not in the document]
kbolashev: Not clear, which is why I print this out explicitly for the user
There's a new PR that allows to mount multiple hooks at the same time that will alleviate this problem though

dagshub/models/model_loaders.py
```diff
@@ -0,0 +1,104 @@
...
+    @property
+    @abstractmethod
+    def model_path(self) -> Path:
+        ...
+
+
+class RepoModelLoader(ModelLoader):
+    def __init__(self, repo_api: RepoAPI, revision: str, path: PurePosixPath):
+        super().__init__(repo_api)
+        self.revision = revision
```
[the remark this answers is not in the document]
kbolashev: Thanks for the catch, made sure it's being used

dagshub/models/model_loaders.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def _eager_load(self, download_dest: Path) -> Path:
+        remote_path = PurePosixPath(f"s3:/{self.repo_api.repo_name}") / self.path
+        local_path = download_dest / self.model_path
+        self.repo_api.download(remote_path, local_path=local_path)
+        return local_path
+
+    @property
+    def model_path(self) -> Path:
+        return Path(".dagshub") / "storage" / "s3" / self.repo_api.repo_name / self.path
```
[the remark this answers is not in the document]
kbolashev: Not really easy to track down, and this is a bit of a band-aid before we redo the paths for all of the storage in `install_hooks()`.
Right now the model_path is this to be consistent with `install_hooks`

dagshub/models/model_loaders.py
```diff
@@ -0,0 +1,104 @@
...
+        local_path = download_dest / self.model_path
+        self.repo_api.download(remote_path, local_path=local_path)
+        return local_path
+
+    @property
+    def model_path(self) -> Path:
+        return Path(self.path)
+
+
+class DagsHubStorageModelLoader(ModelLoader):
```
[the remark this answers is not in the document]
kbolashev: Yea, I guess it should, good catch

dagshub/models/model_locator.py
kbolashev: forgot to delete it.
Left to do: mlflow

dagshub/models/model_locator.py
kbolashev: For the future.
The idea is to actually redownload the model if there's a newer version in the repo (unless a revision was specified).
If we download the model into the hash of the dir, we can do that "easily" by separating the model versions every time.

I didn't really consider how this will work with e.g. buckets, so this is for now left as a "do this in the future"

dagshub/models/model_locator.py
kbolashev: Honestly I'm not doing model imports pretty much anywhere in the code, so if you want to codify this preference into some style guide, I can change it.
Right now this will only result in weird inconsistencies

dagshub/models/model_locator.py
kbolashev: It gets turned into `RepoAPI`, skipping both of the IFs: https://github.com/DagsHub/client/blob/aed3677a7c62bd36d2a599c8447ee166ef54aa9d/dagshub/models/model_locator.py#L82

The `RepoAPI` constructor specifically works with `user/repo` format

dagshub/models/model_locator.py
kbolashev: I don't want to reuse the token cache location for this, because it uses kind of an "AppData" directory, that isn't supposed to be used for file storage
These models will be potentially very big, so they could grow in size a lot, so I want it to be discoverable by users.
We're already using `~/dagshub` for data engine files, so it is reusing already existing path in a way.

dagshub/models/model_locator.py
kbolashev: Yea, that's true, and right now it's not consistent with other ways we use storages.
This is the format that we'll be using later when we reimplement install_hooks pathing, so the hope is that it will be consistent at that point and they wouldn't need to look anything up.

The dir tree after that change should look like this basically:
```
.
|-- README.md
|-- dagshub_storage
|-- my_integrated_bucket
`-- src
```

and the path for the model should be intuitive

dagshub/models/model_locator.py
kbolashev: Yes, it's intended to always be a directory.
We obviously can't stop users at pointing at files, but the way HuggingFace works is that you point at dirs

dagshub/models/model_locator.py
kbolashev: Ah, I used it in two places, but later removed one of those (the one with the specified path).
Inlined it, thanks

tests/model_loading/test_download.py
kbolashev: Forgot to remove it, was a stub file before I started writing tests, then at some point I realized that there's not really a lot of point to testing the downloading, and only the location is the part that needed to be tested really

kbolashev:
[Loading models from the client](https://trello.com/c/FnFEXiVf/1969-loading-models-from-the-client)

dagshub/models/model_loaders.py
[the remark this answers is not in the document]
simonlsk: Ok, I meant easy for you to track down the band-aids when you redo the paths?

dagshub/models/model_locator.py
simonlsk: But kind of same comment as before, the different local paths aren't really managed anywhere, and they're spread inline, so I was implying that I expected to see some DRY for how the path is built. This would allow later changes to be less painful. Not critical for now anyway.

simonlsk (approved):

dagshub/models/model_loaders.py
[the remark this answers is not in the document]
kbolashev: I remember them, and the refactor is supposed to be done soon, so I hope I won't forget.
I'll try to DRY it then, maybe in RepoAPI

PR #450, opened 2024-03-23
Add storage convenience features for mounting and syncing with rclone

deanp70:


dagshub/storage/rclone.py
```diff
@@ -0,0 +1,238 @@
...
+    """
+    Initializes or updates the Rclone configuration for a DAGsHub repository.
+
+    :param repo_owner: The owner of the repository. This is used to create a unique section in the Rclone configuration.
+    :param conf_path: Optional. The path to the Rclone configuration file. If not provided, the default path is used.
+    :param update: Optional. A boolean flag indicating whether to update the configuration if it already exists.
+    Defaults to False.
+    :param verbose: Optional. A boolean flag that controls the output of the function. If True, the function will
+    print messages about its operation.
+    :return: The absolute path to the Rclone configuration file.
```
kbolashev, line 68:
```suggestion
    :return: Name of the remote for rclone + The absolute path to the Rclone configuration file.
```

dagshub/storage/rclone.py
```diff
@@ -0,0 +1,238 @@
...
+        remote_name = f"dagshub-{repo_owner}"
+    else:
+        remote_name = "dagshub"
+
+    config[remote_name] = {
+        "type": "s3",
+        "provider": "Other",
+        "access_key_id": token,
+        "secret_access_key": token,
+        "endpoint": f"https://dagshub.com/api/v1/repo-buckets/s3/{repo_owner}",
```
kbolashev, line 98:
```suggestion
        "endpoint": f"{dagshub.common.config.host}/api/v1/repo-buckets/s3/{repo_owner}",
```

dagshub/storage/rclone.py
```diff
@@ -0,0 +1,238 @@
...
+    :param repo: A string in the "<repo_owner>/<repo_name>" format representing the target DAGsHub repository.
+    """
+    # Extract repo_owner and repo_name from repo argument
+    repo_owner, repo_name = repo.split('/')
+
+    # Ensure the repository is configured in Rclone
+    remote_name, conf_path = rclone_init(repo_owner=repo_owner, verbose=False)
+    if not conf_path or not conf_path.exists():
+        print("Failed to configure rclone.")
+        return
```
kbolashev, line 130: This should be a `RuntimeError` inside of `rclone_init` instead of this check probably

dagshub/storage/rclone.py
```diff
@@ -0,0 +1,238 @@
...
+        if current_dir == repo_name:
+            mount_dir = Path("dagshub_storage")
+        else:
+            mount_dir = Path(repo_name) / "dagshub_storage"
+
+    # 2. Configure the rclone conf
+    remote_name, conf_path = rclone_init(repo_owner=repo_owner, verbose=False)
+    if not conf_path or not conf_path.exists():
+        print("Failed to configure rclone.")
+        return
```
kbolashev, line 184: ditto about the error in `rclone_init`

dagshub/storage/rclone.py
```diff
@@ -0,0 +1,238 @@
...
+    """
+    Mounts a DAGsHub repository bucket to a local directory.
+
+    Note: This function is only supported on Linux machines and on macOS via FUSE for macOS (FUSE-T or macFUSE).
+    It may not work as expected on other operating systems due to differences in the handling of filesystem mounts.
+
+    :param repo: The repository in the format "<repo_owner>/<repo_name>". This is used to determine the remote name and
+    mount point.
+    :param cache: Optional. A boolean flag that enables or disables the cache feature. If True, caching is enabled with
+    specific settings.
```
kbolashev, line 163: Which settings?

dagshub/storage/rclone.py
```diff
@@ -0,0 +1,238 @@
...
+        print("Failed to configure rclone.")
+        return
+
+    # 3. Prepare and execute the mount command
+    try:
+        mount_dir.mkdir(parents=True, exist_ok=True)
+    except OSError:
+        print("Encountered OSError, attempting to remount...")
+        unmount_bucket(repo)
+        mount_dir.mkdir(parents=True, exist_ok=True)
```
kbolashev, line 192: Not sure I understand what's going on here. Can you explain what led you to adding this here?
you try to create a directory, if you failed to do it, you try to unmount then create a directory again. What if the user doesn't have permission to create a directory? Unmounting wouldn't do anything there.

dagshub/storage/rclone.py
```diff
@@ -0,0 +1,238 @@
...
+        # Execute the mount command
+        subprocess.run(mount_command, check=True)
+        print(f"Successfully mounted DagsHub Storage in '{repo_name}' to '{mount_dir}'.")
+    except subprocess.CalledProcessError as e:
+        print(f"Failed to mount DagsHub Storage in '{repo_name}': {e}")
+
+
+def unmount_bucket(repo, path=None):
+    """
+    Unmounts a previously mounted DAGsHub repository bucket from the local file system.
```
kbolashev, line 213: Is there any way in rclone to list current mounts?
Right now this basically mounts and then it's forever there (or until reboot I guess) and the user doesn't really know how to unmount it or where to find what's been done.

dagshub/storage/rclone.py
```diff
@@ -0,0 +1,238 @@
...
+    ]
+
+    # Add cache options if needed
+    if cache:
+        mount_command.extend(["--vfs-cache-max-age", "24h"])
+
+    try:
+        # Execute the mount command
+        subprocess.run(mount_command, check=True)
+        print(f"Successfully mounted DagsHub Storage in '{repo_name}' to '{mount_dir}'.")
```
kbolashev, line 206: Need to print out what to do if I want to unmount it

dagshub/storage/rclone.py
```diff
@@ -0,0 +1,238 @@
...
+    if path:
+        # If a custom path is provided, use it as the mount point
+        mount_point = Path(path)
+    else:
+        # Default logic to determine the mount point
+        current_dir = Path(os.getcwd()).name
+        if current_dir == repo_name:
+            mount_point = Path("dagshub_storage")
+        else:
+            mount_point = Path(repo_name) / "dagshub_storage"
```
kbolashev, line 232: Smart move with naming it `dagshub_storage`

kbolashev (changes requested):

dagshub/storage/rclone.py
```diff
@@ -0,0 +1,238 @@
...
+                except subprocess.CalledProcessError as e:
+                    print(f"Failed to install {package}: {e}")
+        else:
+            print("Skipping installation of missing packages.")
+    else:
+        if verbose:
+            print("All packages are installed.")
+
+
+def rclone_init(repo_owner: str, conf_path: Path = None, update=False, verbose=True) -> tuple[str, Path]:
```
kbolashev, line 58: Python 3.8 doesn't allow type annotations like this, need to `from typing import Tuple`
```suggestion
def rclone_init(repo_owner: str, conf_path: Path = None, update=False, verbose=True) -> Tuple[str, Path]:
```

dagshub/storage/rclone.py
deanp70: This is meant to make sure that silent failures don't lead to problems down the line. Maybe it doesn't throw an error, but doesn't create the config file. Recommendations on how to handle that?

kbolashev: I looked through init_rclone again and I'm not sure if these checks would actually hit:
- If the config didn't get written, you should be getting an `OSError` at line 102
- There's single return with an `.absolute()` on it, so there's no way `conf_path` is either None or blank, it's going to be some actual file inside of the FS.

So maybe you can actually just skip this check completely

dagshub/storage/rclone.py
```diff
@@ -0,0 +1,238 @@
...
+    installation commands.
+
+    :param verbose: Optional. A boolean flag that controls the output of the function. If True, the function will
+    print messages about its operation.
+    No parameters are required for this function.
+    """
+    packages = {
+        "rclone": "sudo -v && curl https://rclone.org/install.sh | sudo bash",
+        "fusermount3": "apt install -y fuse3"
+    }
```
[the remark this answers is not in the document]
deanp70: Fixed this as much as possible, sudo with password will fail but I added the manual installation instructions if it fails to automatically install.

dagshub/storage/rclone.py
deanp70: ok then, will remove it entirely. Less code is better

dagshub/storage/rclone.py
```diff
@@ -0,0 +1,257 @@
...
+
+    # Add cache options if needed
+    if cache:
+        mount_command.extend(["--vfs-cache-max-age", "24h"])
+
+    try:
+        # Execute the mount command
+        subprocess.run(mount_command, check=True)
+        log_message(f"Successfully mounted DagsHub Storage in '{repo_name}' to '{mount_dir}'.")
+        log_message(f"To unmount, run `dagshub.storage.unmount(repo={repo}, path={mount_dir})`.")
```
kbolashev, line 225: Need to add quotes around the paths here
```suggestion
        log_message(f"To unmount, run `dagshub.storage.unmount(repo=\"{repo}\", path=\"{mount_dir}\")`.")
```

kbolashev (approved):
If you want this to show up in the client documentation there's some work that still needs to be done, because you need to add the `dagshub.storage` module to some `.rst` file. 
Might have to change some of the docs after that too to change the formatting.
I'll probably do that right now to save your time

deanp70:
Thanks @kbolashev I'd appreciate it. Can I merge this in then?

PR #454, opened 2024-03-25
Data Engine: Add a progress bar for blob downloading

kbolashev:
This PR adds a progress bar for downloading blob fields in a query result, similar to the one already used for files downloading.

The number in the progress bar corresponds to the total number of non-None blob fields, so it might be bigger or smaller than number of datapoints.
LMK if you want to change it and show something else instead (n_fields * n_datapoints, or just n_datapoints).

### DIFF
--- a/dagshub/data_engine/model/query_result.py
+++ b/dagshub/data_engine/model/query_result.py
@@ -1,11 +1,10 @@
 import inspect
 import logging
-import multiprocessing.pool
+from concurrent.futures import as_completed, ThreadPoolExecutor
 from dataclasses import field, dataclass
-from itertools import repeat
 from os import PathLike
 from pathlib import Path
-from typing import TYPE_CHECKING, List, Dict, Any, Optional, Union
+from typing import TYPE_CHECKING, List, Dict, Any, Optional, Union, Tuple
 
 import rich.progress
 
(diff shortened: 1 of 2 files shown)


kbolashev:
[Add a progress bar for blob downloading](https://trello.com/c/S5ioAnQ7/1977-add-a-progress-bar-for-blob-downloading)

dagshub/data_engine/model/query_result.py
```diff
@@ -268,31 +267,49 @@ def get_blob_fields(
...
+
+        progress = get_rich_progress(rich.progress.MofNCompleteColumn())
+        task = progress.add_task("Downloading blobs...", total=len(to_download))
+
+        auth = self.datasource.source.repoApi.auth
+
+        def _get_blob_fn(dp: Datapoint, field: str, url: str, blob_path: Path):
+            blob_or_path = _get_blob(url, blob_path, auth, cache_on_disk, load_into_memory)
+            if isinstance(blob_or_path, str):
+                logger.warning(f"Error while downloading blob for field {field} in datapoint {dp.path}:{blob_or_path}")
```
guysmoilov, line 302: `_get_blob` will print the error? Not raise? What will be the content of `blob_or_path` in this case?

dagshub/data_engine/model/query_result.py
```diff
@@ -268,31 +267,49 @@ def get_blob_fields(
...
+            dp.metadata[field] = blob_or_path
+
+        with progress:
+            with ThreadPoolExecutor(max_workers=num_proc) as tp:
+                futures = [tp.submit(_get_blob_fn, *args) for args in to_download]
+                for f in as_completed(futures):
+                    exc = f.exception()
+                    if exc is not None:
+                        logger.warning(f"Got exception {type(exc)} while downloading blob: {exc}")
```
guysmoilov, line 311: Maybe the future can yield the actual task so that you can print info on the specific datapoint and blob?

dagshub/data_engine/model/query_result.py
```diff
@@ -268,31 +267,49 @@ def get_blob_fields(
...
-            fields = [field.name for field in self.datasource.fields if field.valueType == MetadataFieldType.BLOB]
-
-        for fld in fields:
-            logger.info(f"Downloading metadata for field {fld} with {num_proc} processes")
-            cache_location = self.datasource.default_dataset_location / ".metadata_blobs"
-
-            if cache_on_disk:
-                cache_location.mkdir(parents=True, exist_ok=True)
-
-            blob_urls = list(map(lambda dp: dp._extract_blob_url_and_path(fld), self.entries))
```
guysmoilov, line 280: Delete the now unused `_extract_blob_url_and_path`

dagshub/data_engine/model/query_result.py
```diff
@@ -268,31 +267,49 @@ def get_blob_fields(
...
+        auth = self.datasource.source.repoApi.auth
+
+        def _get_blob_fn(dp: Datapoint, field: str, url: str, blob_path: Path):
+            blob_or_path = _get_blob(url, blob_path, auth, cache_on_disk, load_into_memory)
+            if isinstance(blob_or_path, str):
+                logger.warning(f"Error while downloading blob for field {field} in datapoint {dp.path}:{blob_or_path}")
+            dp.metadata[field] = blob_or_path
+
+        with progress:
+            with ThreadPoolExecutor(max_workers=num_proc) as tp:
```
guysmoilov, line 306: Think carefully about whether the increased concurrency (across values rather than across fields) will make any race conditions

guysmoilov (approved):

dagshub/data_engine/model/query_result.py
kbolashev: This is a leftover of the previous implementation that didn't use the poolexecutor, where I HAD to handle exception in the function.
`blob_or_path` returns a string with the exception output if there was an exception, and that gets set as the value of the blob field.

Honestly probably not the best outcome, so I can change it

dagshub/data_engine/model/query_result.py
kbolashev: Most of this will be handled by the exception handler in `_get_blob_fn`, where the datapoint will be output as well as the field for which it happened.
This exception handler is more of a doomsday handler for really weird errors.

Besides, if `f.exception()` is not None, then that means that `f` doesn't have a result set, so there's no way for me to know which dp it was working on (unless I store the arguments along with the futures)

dagshub/data_engine/model/query_result.py
kbolashev: ty!!

dagshub/data_engine/model/query_result.py
kbolashev: Dictionaries are thread-safe, and I'm going from a pure [collect all blobs -> set them in one thread] to a concurrent get-and-set.
There shouldn't be any race conditions because there's no way for us to be setting the same key two times on a datapoint, it's always going to be unique key every time

dagshub/data_engine/model/query_result.py
kbolashev: to answer the original question: No exception will be raised if we fail getting a blob value

dagshub/data_engine/model/query_result.py
guysmoilov: OK, if you thought about it carefully

guysmoilov (approved):

PR #458, opened 2024-03-28
Add a dagshub.colab.login command

kbolashev:
This PR adds a `dagshub.colab.login()` command, that leads user through a different auth flow + creates a `<user>/dagshub-drive` repository that will be used for colab-related storage.

### DIFF
--- a/dagshub/auth/oauth.py
+++ b/dagshub/auth/oauth.py
@@ -12,7 +12,7 @@ from dagshub.common import config, rich_console
 logger = logging.getLogger(__name__)
 
 
-def oauth_flow(host: str, client_id: Optional[str] = None) -> OAuthDagshubToken:
+def oauth_flow(host: str, client_id: Optional[str] = None, referrer: Optional[str] = None) -> OAuthDagshubToken:
     """
     Initiate the OAuth 2.0 flow for obtaining an access token.
 
@@ -20,6 +20,7 @@ def oauth_flow(host: str, client_id: Optional[str] = None) -> OAuthDagshubToken:
         host (str): The URL of the OAuth provider.
         client_id (Optional[str], optional): The client ID used for authentication.
             If not provided, it will use the default client ID from the configuration.
+        referrer (Optional[str], optional): For custom referral
 
     Returns:
         Dict: A dictionary containing the obtained access token.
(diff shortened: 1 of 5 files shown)


dagshub/auth/oauth.py
```diff
@@ -31,8 +32,10 @@ def oauth_flow(host: str, client_id: Optional[str] = None) -> OAuthDagshubToken:
     state = uuid.uuid4()
     middle_man_request_id = hashlib.sha256(uuid.uuid4().bytes).hexdigest()
     auth_link = (
-        f"{dagshub_url}/authorize?state={state}&client_id={client_id}" f"&middleman_request_id={middle_man_request_id}"
+        f"{dagshub_url}/authorize?state={state}&client_id={client_id}&middleman_request_id={middle_man_request_id}"
```
guysmoilov, line 35: 1. How did this even work\compile previously?
2. I think you can get rid of the surrounding parens now

dagshub/colab/login.py
```diff
@@ -0,0 +1,33 @@
...
+
+    username = get_user_of_token(token)
+
+    colab_repo = RepoAPI(f"{username}/{COLAB_REPO_NAME}")
+    try:
+        colab_repo.get_repo_info()
+    except RepoNotFoundError:
+        create_repo(COLAB_REPO_NAME)
+    log_message(f"Repository {colab_repo.full_name} is ready for use with Colab. Link to the repository:")
+    print(colab_repo.repo_url)
```
guysmoilov, line 32: Why is one of them `log_message` and the other `print`?

guysmoilov (approved):

dagshub/auth/oauth.py
kbolashev: It was an artifact of the formatter, but the syntax is valid Python and it just sticks them together
you can try this in a REPL:
```
>>> a = "aaa" "bbb"
>>> a
'aaabbb'
```

What happened here is at some point the formatter split it into two strings, but then later when the line length got raised to 120, instead of sticking them together, it just put them on a single line.

The parens I can't get rid of, putting the string inline with auth_link goes over the line length, so the formatter stuck in the parens

dagshub/colab/login.py
kbolashev: The colab/ipython printer works inconsistently with the URLs and not always makes them clickable, so whenever a URL is printed that we want users to be able to click AND we know that it will probably show up in colab, we use regular print.

Other examples:
https://github.com/DagsHub/client/blob/0c2d1ff58037fc0b897d3f9e8d6609ded4c477c9/dagshub/auth/oauth.py#L51
https://github.com/DagsHub/client/blob/bf3832a0350ab9400ee50ad2b25e1da9ba2f8da6/dagshub/data_engine/model/datasource.py#L841

Honestly I can probably replace the log_message with a print for consistency sake.

kbolashev: log_message prints to rich's console which does neat highlights for numbers/urls on some occasions, so I thought to use it here for neater output

PR #461, opened 2024-04-08
Data Engine: Add document blob category

kbolashev:
# Spec:
This PR allows for a, hopefully, better UX for cases when very big string values are used.
The Data Engine backend supports strings at maximum 512 length, and trying to upload a file bigger than that leads to an error from the backend.

As a workaround, with this PR, the client instead uploads big strings as a blob with a `document` tag applied.
Since we can't change types for existing uploaded metadata, this is only valid for adding new fields of metadata in a single batch.

### Validation:
The metadata upload now goes through client-side validation, that will hopefully mitigate making mistakes.
Validation rules implemented with this PR:
- Uploading metadata of different types in a batch is not allowed
- Existing string fields can't have more than 512 symbols in a metadata

### Transforms:
After the metadata validation, there is also now a transform step, that we'll be able to add more transforms later to.
Right now the only transform implemented is the one for the document field:
- If the string field **is new** and has a longest value longer than 512 symbols, all of the values get turned into blobs with UTF-8 encoding, and the field is marked with the `document` tag.

### Miscelaneous UX around the document fields:
- Document fields get autodownloaded on queries and decoded back with UTF-8, so users shouldn't notice that the data is stored as blobs (except for the fact that it's being cached on-disk)
- Same goes for updating the metadata. Putting a string value into a document field "automagically" converts it into a blob. Tested manually for both datapoint updates and metadata context updates, plus there's a test added that checks that it's being transformed into a blob


dagshub/data_engine/model/metadata/validation.py
```diff
@@ -0,0 +1,100 @@
...
+class StringFieldValueTooLongError(Exception):
+    def __init__(self, field_name: str, value: str):
+        self.field_name = field_name
+        self.value = value
+
+    def __str__(self):
+        return (
+            f"Trying to insert a value into a string field {self.field_name} "
+            f"that is longer than the maximum allowed length of {MAX_STRING_FIELD_LENGTH}.\n"
+            "If you need to upload large text, use a document field instead.\n"
```
guysmoilov, line 47: How do I use that?
A code sample would be great

guysmoilov (approved):
I hope I did a good enough job reviewing, looked at pretty much everything but am tired and not going to be around tomrrow - and I would like to ship this to Yair tomorrow.

PR #462, opened 2024-04-08
Data engine: Add an `as_hf_dataset` to QueryResult

kbolashev:
# Spec

This PR adds an `as_hf_dataset` function to the `QueryResult` that allows users to use Data Engine data as a HuggingFace dataset.

Functionality:
- Downloads datapoints and sets the path to the path of the datapoint
- Downloads all blob fields and sets the values to their path
- Gets rid off the autogenerated datapoint_id/dagshub_url fields

### DIFF
--- a/dagshub/data_engine/model/query_result.py
+++ b/dagshub/data_engine/model/query_result.py
@@ -19,7 +19,7 @@ from dagshub.data_engine.annotation.voxel_conversion import (
     add_ls_annotations,
 )
 from dagshub.data_engine.client.models import DatasourceType
-from dagshub.data_engine.model.datapoint import Datapoint, _get_blob
+from dagshub.data_engine.model.datapoint import Datapoint, _get_blob, _generated_fields
 from dagshub.data_engine.client.loaders.base import DagsHubDataset
 from dagshub.data_engine.voxel_plugin_server.utils import set_voxel_envvars
 from dagshub.data_engine.dtypes import MetadataFieldType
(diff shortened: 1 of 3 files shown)


dagshub/data_engine/model/query_result.py
```diff
@@ -297,8 +343,8 @@ def get_blob_fields(
         auth = self.datasource.source.repoApi.auth
 
         def _get_blob_fn(dp: Datapoint, field: str, url: str, blob_path: Path):
-            blob_or_path = _get_blob(url, blob_path, auth, cache_on_disk, load_into_memory)
-            if isinstance(blob_or_path, str):
+            blob_or_path = _get_blob(url, blob_path, auth, cache_on_disk, load_into_memory, path_format)
+            if isinstance(blob_or_path, str) and path_format != "str":
```
guysmoilov, line 347: Any reason to use this unreliable error handling instead of raising errors from `_get_blob`?

dagshub/data_engine/model/query_result.py
```diff
@@ -227,6 +229,43 @@ def keypairs(keys):
...
+
+        df = self.dataframe
+
+        if download_datapoints:
+            # Do the same for the actual datapoint files, changing the path
+            new_paths = []
+            self.download_files()
+            for dp in df["path"]:
+                new_paths.append(
+                    str(self.datasource.default_dataset_location / self.datasource.source.source_prefix / dp)
```
guysmoilov, line 257: Could be nice to let the user specify a non default download path and pass it on to `download_files`

guysmoilov (approved):
Looks good overall, I have suggestions.
Did you find out what the deal is with splits? Will we handle them correctly?

kbolashev:
Note to self: Add a download_path argument here, it will be very useful

dagshub/data_engine/model/query_result.py
kbolashev: Done

kbolashev:
I didn't really look into splits that well, but it doesn't make sense for us to "try" to support them, unless we have splits as a data engine citizen.
Right now the split behavior is reproducible by adding a `split` metadata field and filtering on that. Since this is a function of a QueryResult, any kind of filter can be applied before loading, so that's kind of the same behavior as the splits

PR #463, opened 2024-04-14
Data Engine: Autolog to MLflow on datasource queries

kbolashev:
# Implemented in this PR:
- Added a link to open the query in the gallery view
- Added autologging **only on ds.all()** of queries to MLflow
- The logging happens in a parallel thread so as to not interfere with the querying
- Querying works with multiple datasources, correctly saving the query for each ds.
- A new run is created in the last modified experiment per datasource that's being queried, all the queries are then logged into that run.

# Left to implement:
- RIght now the artifacts all get logged with the same name, so they are being overwritten every time. I need to add a tracker for how many logs has been saved.
- Probably would be a good idea to have the autologging togglable in some ways, because the runs created are not being closed and are going to pollute the environment, so I expect advanced users to be able to turn this off.

# Caveats/bugs:
- Closing a run doesn't work if there's a logging session going on to another repo's MLflow. Possible solution - if the current datasource is not in the active run's repo, log in foreground instead of background.

### DIFF
--- a/dagshub/data_engine/model/datasource.py
+++ b/dagshub/data_engine/model/datasource.py
@@ -750,7 +750,8 @@ class Datasource:
         if not is_mlflow_installed:
             return
         # Run ONLY if there's an active run going on
-        if mlflow.active_run() is None:
+        active_run = mlflow.active_run()
+        if active_run is None:
             return
         source_name = self.source.name
 
@@ -758,9 +759,7 @@ class Datasource:
         uuid_chunk = str(uuid.uuid4())[-4:]
 
         artifact_name = f"autolog_{source_name}_{now_time}_{uuid_chunk}.dagshub.json"
-        threading.Thread(
-            target=self.log_to_mlflow, kwargs={"artifact_name": artifact_name, "run": mlflow.active_run()}
-        ).start()
+        threading.Thread(target=self.log_to_mlflow, kwargs={"artifact_name": artifact_name, "run": active_run}).start()
 
     def log_to_mlflow(
         self, artifact_name=DEFAULT_MLFLOW_ARTIFACT_NAME, run: Optional["mlflow.entities.Run"] = None
--- a/dagshub/data_engine/model/datasource.py
+++ b/dagshub/data_engine/model/datasource.py
@@ -6,8 +6,8 @@ import math
 import os.path
 import threading
 import time
+import uuid
 import webbrowser
-from collections import defaultdict
 from contextlib import contextmanager
 from dataclasses import dataclass, field
 from os import PathLike
@@ -66,9 +66,6 @@ else:
 logger = logging.getLogger(__name__)
 
 DEFAULT_MLFLOW_ARTIFACT_NAME = "datasource.dagshub.json"
-
-_autolog_counters: Dict[str, int] = defaultdict(int)
-_autolog_counter_lock = threading.Lock()
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


kbolashev:
Changes to be done:
- Log ONLY to the active run, no point in creating a run on the datasource's repo as it turns out
- Come up with a better naming in these cases (probably log the name of the repo+datasource+index), would be good if the index could be gotten from mlflow also.

kbolashev:
Changed it so it logs to the current one.
The format is `autolog_<datasource_name>_<#>.dagshub.json`. Decided that adding the repo is a bit too verbose

dagshub/data_engine/model/datasource.py
```diff
@@ -742,6 +749,19 @@ def save_dataset(self, name: str) -> "Datasource":
...
+    def _autolog_mlflow(self):
+        if not is_mlflow_installed:
+            return
+        # Run ONLY if there's an active run going on
+        if mlflow.active_run() is None:
+            return
+        source_name = self.source.name
+        with _autolog_counter_lock:
+            counter_value = _autolog_counters[source_name]
+            _autolog_counters[source_name] += 1
```
guysmoilov, line 761: Couple issues here:
1. Race condition on multiprocessing, artifacts will override each other
2. If resuming an existing run and logging more to it, the counters don't persist.

Overall, it might be more sane to just use a UUID with a human readable timestamp format...
(I would stick to the timestamp formats we use in our UI when picking versions to compare, which IIRC is RFC3339)
(And some random bits before the timestamp to make it a "UUID" )

dagshub/data_engine/model/datasource.py
```diff
@@ -742,6 +749,19 @@ def save_dataset(self, name: str) -> "Datasource":
...
+        artifact_name = f"autolog_{source_name}_{counter_value}.dagshub.json"
+        threading.Thread(target=self.log_to_mlflow, kwargs={"artifact_name": artifact_name}).start()
```
guysmoilov, line 763: Would be wise to pass along the active run as a parameter

guysmoilov: And, are you very sure that `log_to_mlflow` is safe to do async?
The datasource object won't be modified, or otherwise be made into garbage by the time `log_to_mlflow` gets executed?

guysmoilov (changes requested):

dagshub/data_engine/model/datasource.py
kbolashev: The datasource object is deep copied on every modification, so `log_to_mlflow`is guaranteed to be working with the same object

dagshub/data_engine/model/datasource.py
```diff
@@ -756,11 +753,14 @@ def _autolog_mlflow(self):
...
-            _autolog_counters[source_name] += 1
-        artifact_name = f"autolog_{source_name}_{counter_value}.dagshub.json"
-        threading.Thread(target=self.log_to_mlflow, kwargs={"artifact_name": artifact_name}).start()
+
+        now_time = datetime.datetime.now().strftime("%Y-%m-%dT%H-%M-%S")  # Not ISO format to make it a valid filename
+        uuid_chunk = str(uuid.uuid4())[-4:]
+
+        artifact_name = f"autolog_{source_name}_{now_time}_{uuid_chunk}.dagshub.json"
+        threading.Thread(
+            target=self.log_to_mlflow, kwargs={"artifact_name": artifact_name, "run": mlflow.active_run()}
```
guysmoilov, line 762: Might want to reuse the `active_run() from line 753

guysmoilov (approved):

PR #467, opened 2024-04-21
Implement thumbnails

TalMalka123:
Add a function to tag metadata as a thumbnail/ remove thumbnail tag.

### DIFF
--- a/dagshub/data_engine/model/metadata_field_builder.py
+++ b/dagshub/data_engine/model/metadata_field_builder.py
@@ -1,6 +1,6 @@
 import dataclasses
 import logging
-from typing import TYPE_CHECKING, Type, Union, Set
+from typing import TYPE_CHECKING, Type, Union, Set, Optional, Literal
 
 from dagshub.data_engine.client.models import MetadataFieldSchema
 from dagshub.data_engine.dtypes import DagshubDataType, MetadataFieldType, ReservedTags, ThumbnailType
@@ -80,7 +80,7 @@ class MetadataFieldBuilder:
         self._set_or_unset(ReservedTags.ANNOTATION.value, is_annotation)
         return self
 
-    def set_thumbnail(self, thumbnail_type: Union[ThumbnailType, str, None] = None,
+    def set_thumbnail(self, thumbnail_type: Optional[Literal["video", "audio", "image", "pdf", "text", "csv"]] = None,
                       is_thumbnail: bool = True) -> "MetadataFieldBuilder":
         """
         Mark or unmark the field as thumbnail field, with the specified thumbnail type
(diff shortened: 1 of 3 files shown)


dagshub/data_engine/model/metadata_field_builder.py
```diff
@@ -80,6 +80,57 @@ def set_annotation(self, is_annotation: bool = True) -> "MetadataFieldBuilder":
...
+        elif thumbnail_type == ThumbnailType.IMAGE:
+            tag = ReservedTags.IMAGE_THUMBNAIL_VIZ
+        elif thumbnail_type == ThumbnailType.PDF:
+            tag = ReservedTags.PDF_THUMBNAIL_VIZ
+        elif thumbnail_type == ThumbnailType.TEXT:
+            tag = ReservedTags.TEXT_THUMBNAIL_VIZ
+        elif thumbnail_type == ThumbnailType.CSV:
+            tag = ReservedTags.CSV_THUMBNAIL_VIZ
+        else:
+            raise ValueError("Invalid thumbnail type")
```
kbolashev, line 110: Let the user know what's the invalid type + what are the valid types

dagshub/data_engine/model/metadata_field_builder.py
```diff
@@ -80,6 +80,57 @@ def set_annotation(self, is_annotation: bool = True) -> "MetadataFieldBuilder":
...
+
+        self._set_or_unset_thumbnails(tag, True)
+        return self
+
+    def _set_or_unset_thumbnails(self, type_tag, is_thumbnail):
```
kbolashev, line 115: The `is_thumbnail` arg in this function is only set to True, so the code in 131 will never run. 

I think you should remove the arg and inline the function back into `set_thumbnail`

dagshub/data_engine/model/metadata_field_builder.py
```diff
@@ -80,6 +80,57 @@ def set_annotation(self, is_annotation: bool = True) -> "MetadataFieldBuilder":
         self._set_or_unset(ReservedTags.ANNOTATION.value, is_annotation)
         return self
 
+    def set_thumbnail(self, thumbnail_type: Union[ThumbnailType, None] = None,
+                      is_thumbnail: bool = True) -> "MetadataFieldBuilder":
```
kbolashev, line 84: What happens when I call it like this:
```python
set_thumbnail("image")
```
?

Honestly I think having to import another enum for this is a bit unwieldy for user side.
You can probably allow strings in thumbnail_type and then in the beginning of the function (after the is None check) do:
```python
if isinstance(thumbnail_type, str):
    thumbnail_type = ThumbnailType(thumbnail_type)
```

dagshub/data_engine/model/metadata_field_builder.py
```diff
@@ -80,6 +80,57 @@ def set_annotation(self, is_annotation: bool = True) -> "MetadataFieldBuilder":
...
+        """
+        Mark or unmark the field as thumbnail field, with the specified thumbnail type
+        """
```
kbolashev, line 87: Add docs for the arguments

kbolashev (changes requested):

dagshub/data_engine/model/metadata_field_builder.py
TalMalka123: No its not true, in line 89 I call it with false
image

dagshub/data_engine/model/metadata_field_builder.py
```diff
@@ -100,6 +160,12 @@ def _remove_tag(self, tag: str):
         except ValueError:
             logger.warning(f"Tag {tag} doesn't exist on the field, nothing to delete")
 
+    def _remove_tags(self, tags: Set[str]):
+        if self.schema.tags is None:
+            return
+        for t in tags:
+            self._remove_tag(t)
```
kbolashev, line 167: Change the signature
```suggestion
    def _remove_tags(self, *tags: str):
        if self.schema.tags is None:
            return
        for t in tags:
            self._remove_tag(t)
```

And in the usages removes the curly braces

kbolashev (approved):

PR #475, opened 2024-05-02
Add a monkey patch to make MLflow logging functions not throw after failed retries

kbolashev:
Customer request: make it so MLflow logging functions failing doesn't crash the program.
For now it's done by making MLflow functions fail "silently" (failure gets printed out in the log).

For this I'm basically patching every function in the `mlflow` package to add a try/catch.
The try/catch bubbles up exceptions until the topmost call in the mlflow package, so all handling inside of mlflow should be running correctly.

I'm also patching all functions in `mlflow.MlflowClient` class.
The functions that return a context manager I'm retrying for 15 minutes instead, because they have to return something. Maybe I could be returning a mock context manager that no-ops.

```python
with mlflow.start_run():
    # Code here
```

I don't think this is in general a good approach, because, for example, silencing an error on a function that's supposed to be returning some value, leads to the function returning nothing. This could break something down the way in the user code, for example:

```python
experiment = mlflow.get_experiment("My Experiment") # This fails
mlflow.delete_experiment(experiment.id) # Now here there's no experiment
```

IMO, it would be better if users wrap their code in a `try/except MlflowException` and achieve the same result without the potential for weird subtle breakages like this.

Example code that this was tested with:
```python
import mlflow
from dagshub.mlflow import patch_mlflow

patch_mlflow()

with mlflow.start_run():
    mlflow.log_metric("logged_metric_1", 1)
    time.sleep(10) # Turn off server here
    mlflow.log_metric("unlogged_metric_1", 2)
    time.sleep(15) # Turn server back on here
    mlflow.log_metric("logged_metric_2", 3)
```

The result:
image

The running script printed out the exception for the unlogged metric log

### DIFF
--- a/dagshub/mlflow/patch.py
+++ b/dagshub/mlflow/patch.py
@@ -41,17 +41,14 @@ _mlflow_client_functions = [
 
 _default_patch_functions = _top_level_functions + _mlflow_client_functions
 
-_default_patch_classes = ["mlflow.tracking.client.MlflowClient"]
-
 _default_guaranteed_raises = ["log_model"]
 
 
 class MlflowMonkeyPatch:
     context_manager_functions = ["start_run"]
 
-    def __init__(self, funcs_to_patch: List[str], classes_to_patch: List[str], guaranteed_raises: List[str]):
+    def __init__(self, funcs_to_patch: List[str], guaranteed_raises: List[str]):
         self.funcs_to_patch = set(funcs_to_patch)
-        self.classes_to_patch = set(classes_to_patch)
         # Functions that guarantee to raise an exception if they are in the stack
         # Example: all mlflow.<framework>.log_model functions underneath use log_artifact, but we might want to raise them
         self.functions_with_guaranteed_raises = set(guaranteed_raises)
@@ -141,7 +138,7 @@ class MlflowMonkeyPatch:
                 setattr(module, attr_name, self._wrap_func(attr))
                 continue
 
-            if is_class and full_attr_name in self.classes_to_patch:
+            if is_class:
                 for class_attr_name in dir(attr):
                     if class_attr_name.startswith("_"):
                         continue
@@ -197,7 +194,7 @@ _global_mlflow_patch: Optional[MlflowMonkeyPatch] = None
 _patch_mutex = threading.Lock()
 
 
-def _patch_mlflow(funcs_to_patch: List[str], classes_to_patch: List[str], guaranteed_raises: List[str]):
+def _patch_mlflow(funcs_to_patch: List[str], guaranteed_raises: List[str]):
     import mlflow
 
     global _global_mlflow_patch
@@ -212,7 +209,7 @@ def _patch_mlflow(funcs_to_patch: List[str], classes_to_patch: List[str], guaran
             "Call dagshub.mlflow.unpatch_mlflow() to undo"
         )
 
-        patch = MlflowMonkeyPatch(funcs_to_patch, classes_to_patch, guaranteed_raises)
+        patch = MlflowMonkeyPatch(funcs_to_patch, guaranteed_raises)
         patch.monkey_patch_module(mlflow)
         _global_mlflow_patch = patch
 
@@ -220,22 +217,19 @@ def _patch_mlflow(funcs_to_patch: List[str], classes_to_patch: List[str], guaran
 def _resolve_patches(
     include: Optional[List[str]],
     exclude: Optional[List[str]],
-    include_classes: Optional[List[str]],
     patch_top_level: bool,
     patch_mlflow_client: bool,
     raise_on_log_model: bool,
-) -> Tuple[List[str], List[str], List[str]]:
+) -> Tuple[List[str], List[str]]:
     """
     Resolve the arguments for patch_mlflow
 
     Returns:
         Tuple of:
         - List of functions that need to be patched
-        - List of classes that need to be patched
         - List of functions that should raise exceptions if they are in the stack
     """
     funcs_to_patch = _default_patch_functions.copy()
-    classes_to_patch = _default_patch_classes.copy()
     guaranteed_raises = _default_guaranteed_raises.copy()
 
     if not patch_top_level:
@@ -247,13 +241,9 @@ def _resolve_patches(
         for f in _mlflow_client_functions:
             if f in funcs_to_patch:
                 funcs_to_patch.remove(f)
-        classes_to_patch.remove("mlflow.tracking.client.MlflowClient")
 
     if include is not None:
         funcs_to_patch = funcs_to_patch + include
-
-    if include_classes is not None:
-        classes_to_patch = classes_to_patch + include_classes
 
     if exclude is not None:
         for f in exclude:
@@ -264,16 +254,15 @@ def _resolve_patches(
         if "log_model" in guaranteed_raises:
             guaranteed_raises.remove("log_model")
 
-    return funcs_to_patch, classes_to_patch, guaranteed_raises
+    return funcs_to_patch, guaranteed_raises
 
 
 def patch_mlflow(
     include: Optional[List[str]] = None,
     exclude: Optional[List[str]] = None,
-    include_classes: Optional[List[str]] = None,
     raise_on_log_model=True,
     patch_top_level=True,
-    patch_mlflow_client=False,
+    patch_mlflow_client=True,
 ):
     """
     Patch MLflow functions, making some of them stop raising exceptions, instead logging them to console.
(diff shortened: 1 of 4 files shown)


guysmoilov:
Interesting, I didn't have the various bookkeeping functions like `start_run` in mind when I asked for this.
For those, it makes more sense to throw an error, since they're more likely to be O(1) operations and less likely to fail in the middle of a days long training run.
I was thinking more of the `log_*` methods which continuously monitor a long lived process.

dagshub/mlflow/patch.py
```diff
@@ -0,0 +1,176 @@
...
+        return
+
+    logger.warning(
+        "Patching MLflow to prevent any MLflow exceptions from being raised. "
+        "Call dagshub.mlflow.unpatch_mlflow() to undo"
+    )
+
+    patch = MlflowMonkeyPatch()
+    patch.monkey_patch_module(mlflow)
+    _global_mlflow_patch = patch
```
guysmoilov, line 169: You need a mutex here and in the unpatch method to avoid race conditions

dagshub/mlflow/patch.py
```diff
@@ -0,0 +1,176 @@
...
+            attr = sys.modules[module_name]
+            for index, attr_name in enumerate(attrs):
+                parent_attr = attr
+                attr = getattr(parent_attr, attr_name)
+                if index == len(attrs) - 1:
+                    setattr(parent_attr, attr_name, unpatched_func)
+
+        self.patched_modules.clear()
+        self.original_func_lookup.clear()
+        _global_mlflow_patch = None
```
guysmoilov, line 147: I don't like the the object is modifying the global flag, breaking encapsulation.
The flag should be set and unset only from the global functions below

dagshub/mlflow/patch.py
```diff
@@ -0,0 +1,176 @@
...
+            return wrap_fn
+
+    def is_function_already_patched(self, func_name: str) -> bool:
+        return func_name in self.original_func_lookup
+
+    def monkey_patch_module(self, module: ModuleType):
+        from mlflow.utils.lazy_load import LazyLoader
+
+        # Add module into set of patched modules, so there's no circular import recursion
+        self.patched_modules.add(module.__name__)
```
guysmoilov, line 77: 1. You should probably use mutexes in these methods that check and modify the shared dict and set of patched methods and modules
2. What if it's already in the set? Shouldn't you skip the rest of this logic?

dagshub/mlflow/patch.py
```diff
@@ -0,0 +1,176 @@
...
+from typing import Callable, Optional, Dict, Set
+
+from tenacity import Retrying, stop_after_delay
+
+logger = logging.getLogger(__name__)
+
+
+class MlflowMonkeyPatch:
+    context_manager_functions = ["start_run"]
+    classes_to_patch = ["mlflow.tracking.client.MlflowClient"]
```
guysmoilov, line 15: As we discussed, these lists should:
1. By default only include the O(n) unimportant logging functions - `log_artifact`, `log_artifacts`, `log_param`, `log_param`, `log_metric`, `log_figure`, `log_image`, `log_dict`, `log_table`, `log_text`
2. Be customizable (more about that below in the comment about `patch mlflow()`

dagshub/mlflow/patch.py
```diff
@@ -0,0 +1,176 @@
...
+
+        self.patched_modules.clear()
+        self.original_func_lookup.clear()
+        _global_mlflow_patch = None
+
+
+_global_mlflow_patch: Optional[MlflowMonkeyPatch] = None
+
+
+def patch_mlflow():
```
guysmoilov, line 153: IMO, there should be a low-level global function that accepts the explicit lists of functions and classes to patch, and a higher level abstract function that accepts user-facing common flags:
1. `patch_artifact_uploads=True`
2. `additional_functions=[]`
3. Anything other high-level that you might think of. e.g. a default-off flag to patch `start_run`

Also, it would be extra nice to call the global patch function from our `dagshub.init(mlflow=True)` function, you can add an additional flag `patch_mlflow=False` by default

dagshub/mlflow/patch.py
```diff
@@ -0,0 +1,176 @@
...
+    patch = MlflowMonkeyPatch()
+    patch.monkey_patch_module(mlflow)
+    _global_mlflow_patch = patch
+
+
+def unpatch_mlflow():
+    if _global_mlflow_patch is None:
+        logger.warning("MLflow should be unpatched already")
+        return
+    _global_mlflow_patch.unpatch()
```
guysmoilov, line 176: Suggestion if it's not too difficult - could you add a context manager that temporarily turns off the patch while inside the context? If I have a specific important call. WDYT?
```python
with dagshub.mlflow.pause_patch():
   mlflow.log_param('super-important', 42)
```

guysmoilov (changes requested):
Suggested changes, please a bit for user feedback before implementing

dagshub/mlflow/patch.py
kbolashev: 1. This is not happening live at user-call-time, it's a one-off function at init level. Don't see a point in trying to make it thread-safe. Putting a lock on the patch/unpatch functions will have the same result.
2. It is being skipped, see line 125. Before I recurse into this function with a new module, I see that it wasn't patched already

dagshub/mlflow/patch.py
kbolashev: This is actually dangerous thread-wise, because I'm patching the whole mlflow module, so any parallel request will also have the patch unapplied.

If you still want this functionality, as an alternative I can add a kwarg to the wrapping function that disables the patch for this function.
```
mlflow.log_param(blabla, dont_ignore_exception=True)
```

The only problem is it kind of prevents code reuse + you have to duplicate the flag in however many calls to mlflow you do

guysmoilov: As we discussed, you could set a thread-local flag that causes the existing hooks to no-op while it's on.
But leaving that out of the current PR.

dagshub/mlflow/patch.py
```diff
@@ -1,20 +1,60 @@
...
+_default_patch_classes = ["mlflow.tracking.client.MlflowClient"]
+
+_default_guaranteed_raises = ["log_model"]
+
 
 class MlflowMonkeyPatch:
     context_manager_functions = ["start_run"]
-    classes_to_patch = ["mlflow.tracking.client.MlflowClient"]
+    # Functions that guarantee to raise an exception if they are in the stack
+    # Example: all mlflow.<framework>.log_model functions underneath use log_artifact, but we might want to raise them
```
guysmoilov, line 52: This comment seems misplaced

dagshub/mlflow/patch.py
```diff
@@ -24,15 +64,18 @@ def _wrap_func(self, fn: Callable) -> Callable:
...
 
             wrap_fn_met = False
             for stack_line in tb:
-                is_wrap_func = stack_line.filename == __file__ and stack_line.name in [
+                # If there is a guaranteed raise in the stack - return False
+                if stack_line.name in self.functions_with_guaranteed_raises:
+                    return False
+                is_wrap_func = stack_line.filename == __file__ and stack_line.name in {
                     "wrap_fn",
                     "context_manager_wrap_fn",
```
guysmoilov, line 77: Actually, IMO it would be wise to make these names much less generic.
Give them something like a `dh_mlflow_` prefix

dagshub/mlflow/patch.py
```diff
@@ -0,0 +1,328 @@
...
+    return funcs_to_patch, classes_to_patch, guaranteed_raises
+
+
+def patch_mlflow(
+    include: Optional[List[str]] = None,
+    exclude: Optional[List[str]] = None,
+    include_classes: Optional[List[str]] = None,
+    raise_on_log_model=True,
+    patch_top_level=True,
+    patch_mlflow_client=False,
```
guysmoilov, line 275: Why is the default here false?

dagshub/mlflow/patch.py
```diff
@@ -0,0 +1,328 @@
...
+        exclude: List of functions you DON'T want to patch.\
+            Use this if you need to make sure, for example, all ``log_artifact``
+            functions raise an exception on failure:
+
+            .. code-block:: python
+
+                patch_mlflow(exclude=["mlflow.log_artifact", "mlflow.log_artifacts"])
+
+        include_classes: If ``include`` has classes other than the MlflowClient,
+            you need to list the class here, otherwise it won't be patched
```
guysmoilov, line 295: I wonder if the `classes_to_patch` flag isn't just adding unneeded complexity?
Couldn't you scan all the classes you find and patch any included functions you find?

dagshub/mlflow/patch.py
```diff
@@ -0,0 +1,328 @@
...
+        raise_on_log_model: If true, patched log calls still raise an exception
+            if they are called from a ``log_model`` function.
```
guysmoilov, line 297:
```suggestion
        raise_on_log_model: If true, patched log calls still raise an exception
            if they are called from a ``log_model`` function.
            This ensures that even if you want to ignore failed calls to ``log_artifact`` or ``log_figure`` etc., 
            failed calls to `log_model` in particular will still raise errors.
            This should work for all supported ML frameworks, as long as the function is named ``log_model``.
```

guysmoilov (changes requested):

dagshub/mlflow/patch.py
kbolashev: Good catch, I set it to False before I did the log_model failure thing and forgot to return it back.
Thank you

dagshub/mlflow/patch.py
kbolashev: The guard for "is this function a wrapper function" also checks that the function file is this particular file.
I'll change it just so the stack traces look better I guess

guysmoilov (approved):

PR #487, opened 2024-06-04
Infra: Utilize fields that are returned from the backend on querying in QueryResult

kbolashev:
## Spec

Right now any function in QueryResult that interacts with the fields uses the Datasource's fields.
This is not correct and breaks the moment user queries additional fields.

Example of a broken interaction that results from this:
```python
old_field = Field("blob_field", alias="blob_field_old", as_of=dateutil.parser.parse("Wed 22 Nov 2023"))
qr = ds.select("*", old_field).all()
qr.get_blob_fields()
```

You would expect both `blob_field` and `blob_field_old` to be downloaded, but because QueryResult lacked the info about the selected fields, `blob_field_old` gets ignored.
This PR fixes this, now `QueryResult` has a `fields` field that is filled in from the GraphQL query.

## Still to do

This PR still doesn't forbid users from trying to update metadata on some of those selected annotations, this is out of scope of this PR but I'll make sure to add it to the task log.


dagshub/data_engine/model/query_result.py
```diff
@@ -110,13 +113,23 @@ def __repr__(self):
 
     @staticmethod
     def from_gql_query(query_resp: Dict[str, Any], datasource: "Datasource") -> "QueryResult":
-        if "edges" not in query_resp:
-            return QueryResult([], datasource)
-        if query_resp["edges"] is None:
-            return QueryResult([], datasource)
+        raw_fields = query_resp.get("selectFields", [])
```
sdafni, line 116: can short to raw_fields = query_resp.get("selectFields", []) or []

PR #493, opened 2024-07-01
Enhancement / Print “accessing as <user>” in client

EvgeniLeonti:
notion ticket: https://www.notion.so/dagshub/Print-accessing-as-user-in-client-92af32ddfce2477ba7184722634fb891?pvs=4

### DIFF
--- a/dagshub/auth/tokens.py
+++ b/dagshub/auth/tokens.py
@@ -231,9 +231,10 @@ class TokenStorage:
             # 500's might be ok since they're server errors, so check only for 400's
             assert not (400 <= resp.status_code <= 499)
             if resp.status_code == 200:
-                assert "login" in resp.json()
-            user = resp.json()
-            return user
+                user = resp.json()
+                assert "login" in user
+                assert "username" in user
+                return user
         except AssertionError:
             return None
 
--- a/dagshub/auth/tokens.py
+++ b/dagshub/auth/tokens.py
@@ -4,7 +4,6 @@ import os
 import threading
 import traceback
 from typing import Optional, Dict, List, Set, Union
-from multiprocessing import AuthenticationError
 
 import yaml
 from httpx import Auth
@@ -132,7 +131,7 @@ class TokenStorage:
             if user is not None:
                 self._print_accessing_as(user)
             else:
-                raise AuthenticationError("Provided DagsHub token is not valid")
+                raise RuntimeError("Provided DagsHub token is not valid")
             return EnvVarDagshubToken(config.token, host)
 
         with self._token_access_lock:
--- a/dagshub/auth/tokens.py
+++ b/dagshub/auth/tokens.py
@@ -127,7 +127,10 @@ class TokenStorage:
 
         host = host or config.host
         if host == config.host and config.token is not None:
-            self._print_accessing_as(config.token, host)
+            user = TokenStorage.get_username_of_token(config.token, host)
+            if user is not None:
+                self._print_accessing_as(user)
+            # todo discuss with Kiril about cases when token is not valid
             return EnvVarDagshubToken(config.token, host)
 
         with self._token_access_lock:
@@ -137,6 +140,7 @@ class TokenStorage:
                 self._known_good_tokens[host] = set()
             good_token_set = self._known_good_tokens[host]
             good_token = None
+            good_user = None
             token_queue = list(sorted(tokens, key=lambda t: t.priority))
 
             for token in token_queue:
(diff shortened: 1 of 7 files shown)


EvgeniLeonti:
@kbolashev i have implemented initial version which works fine, but isn't efficient:
it will make another http request to get the username from the token.

in cases where we call `is_valid_token()`, this new http request can be skipped.

so we have 2 options, leave it as is, or make required adjustments to skip the redundant http call.
in order to achieve the 2nd option, i thought about:
1. adding username to `__token_cache`
2. maintain existing caches to support backwards compatibility (adding username to already existing tokens)

this will require more complicated logic, and i'm not sure the scope of this task covers that.

WDYT?

kbolashev:
I think you can change the signature and name of `is_valid_token` to return the username and get 99% of the way there.
You'll only have to call it for the case of the EnvVar token and that's it.

dagshub/auth/tokens.py
```diff
@@ -222,10 +231,23 @@ def is_valid_token(token: Union[str, Auth, DagshubTokenABC], host: str) -> bool:
             # 500's might be ok since they're server errors, so check only for 400's
             assert not (400 <= resp.status_code <= 499)
             if resp.status_code == 200:
```
kbolashev, line 233: A 500 will not return the dictionary, so probably should change it to this now:
```suggestion
            assert resp.status_code == 200
```

dagshub/auth/tokens.py
```diff
@@ -222,10 +231,23 @@ def is_valid_token(token: Union[str, Auth, DagshubTokenABC], host: str) -> bool:
...
+                assert "login" in user
+                assert "username" in user
+                return user
         except AssertionError:
-            return False
+            return None
+
+    @staticmethod
+    def is_valid_token(token: Union[str, Auth, DagshubTokenABC], host: str) -> bool:
+        """
```
kbolashev, line 243: Great job on still leaving this! I honestly forgot it's used in a couple other places 🙃

dagshub/auth/tokens.py
```diff
@@ -281,6 +303,24 @@ def _store_cache_file(self):
             logger.error(f"Error while storing DagsHub token cache: {traceback.format_exc()}")
             raise
 
+    def _print_accessing_as(self, user):
```
kbolashev, line 306:
```suggestion
    def _print_accessing_as(self, user: Dict):
```

dagshub/auth/tokens.py
```diff
@@ -281,6 +303,24 @@ def _store_cache_file(self):
...
+        It called after successful token validation.
+
+        """
+        if self._accessing_as_was_printed:
+            return
+
+        username = user.get("username")
+        if username is None:
+            # todo handle error ? it should be defined
+            return
```
kbolashev, line 319: Something probably went SUPER wrong if we got the response from the API, but the structure of the dictionary is different

I say just skip the whole check and just do a regular dict access. If such a weird case happens, it's probably a backend error and there's no point in trying to handle it in the client
```suggestion
        username = user["username"]
```

tests/common/test_determine_repo.py
```diff
@@ -2,13 +2,15 @@
...
 from typing import Generator, TypeVar
+import respx
 
 import pytest
 import pytest_git
 
 from dagshub.common.determine_repo import determine_repo, parse_dagshub_remote
 import dagshub.common.config
 from dagshub.common.errors import DagsHubRepoNotFoundError
+from tests.dda.test_tokens import valid_token_side_effect
```
kbolashev, line 13: Maybe move this to `tests.util` or someplace like that.
Accessing across the test modules like that looks weird

kbolashev (changes requested):
Looks great!

I left a bunch of minor things that need to be changed, but the logic is great.

dagshub/auth/tokens.py
```diff
@@ -4,6 +4,7 @@
 import threading
 import traceback
 from typing import Optional, Dict, List, Set, Union
+from multiprocessing import AuthenticationError
```
kbolashev, line 7: Very weird module to import from, either create a new error class, or use RuntimeError

Examples: https://github.com/DagsHub/client/blob/master/dagshub/data_engine/model/errors.py

kbolashev (changes requested):

dagshub/auth/tokens.py
EvgeniLeonti: that was weird to me as well, took it from `dagshub/streaming/filesystem.py`

kbolashev (approved):
Congrats on the first feature PR for the client!!! 🥳

PR #494, opened 2024-07-01
Added get_predictions on QueryResult

jinensetpal:
Created a `get_predictions` function that:
1. Loads a model registered with DagsHub-hosted MLFlow
2. Locally performs inference on it and either appends to a column, and/or returns it for further use.


kbolashev:
Can you redo the linter/long-lines commits with line length set to 120, please?
Too many changes, which tells me that it was probably done with 80.

What are you using to format?
If it's black, rerun it with `--line-length 120`

I hope to move to `pyproject.toml` one day so I can keep this consistent :)

jinensetpal:
On it!

jinensetpal:
Interesting note, black does not format comments if they are over the length limit, but flake8 raises issues. What's the preferred approach?

kbolashev:
> Interesting note, black does not format comments if they are over the length limit, but flake8 raises issues. What's the preferred approach?

I just line break them manually, so it makes sense logically.
I switched from black+flake8 to ruff recently, it's a lot faster so I usually get complaints about long line lengths immediately

dagshub/data_engine/model/query_result.py
```diff
@@ -623,22 +694,63 @@ def visualize(self, visualizer: Literal["dagshub", "fiftyone"] = "dagshub", **kw
...
+        model_repo=None,
+        model_pre_hook=lambda x: x,
+        model_post_hook=None,
+        model_batch_size=1,
+        model_metadata_column="annotation",
+        use_remote_backend=False,
     ) -> Optional[str]:
         """
         Sends all the datapoints returned in this QueryResult to be annotated in Label Studio on DagsHub.
+        Alternatively, uses MLFlow to automatically label datapoints.
```
kbolashev, line 715: 1) Split this into a different function, something like `annotate_with_mlflow`. I understand that this name of a function is kind of too generic, but making it have two wildly different behaviors depending on args is imo a bad pattern.
2) Add typehints to the args (you can add them to this one too if you want)
3) When you move it to a separate function, you can probably drop all the `model_` prefixes

dagshub/data_engine/model/query_result.py
```diff
@@ -401,6 +404,74 @@ def download_binary_columns(
...
+            version: (optional, default: 'latest') version of the model in the mlflow registry
+            pre_hook: (optional, default: identity function) function that runs before datapoint is sent to the model
+            post_hook: (optional, default: identity function) function that converts mlflow model output
+            to the desired format
+            batch_size: (optional, default: 1) function that sets batch_size
+            metadata_column: (optional, default: 'prediction') write prediction results to metadatalogged in data engine.
+            If None, returns predictions.
+            return_predictions: (optional, default: True) returns predictions logged
+        Returns:
+            The URL of the created Label Studio workspace
```
kbolashev, line 440: Doesn't look like it does that

dagshub/data_engine/model/query_result.py
```diff
@@ -401,6 +404,74 @@ def download_binary_columns(
...
+    ):
+        """
+        Sends all the datapoints returned in this QueryResult as prediction targets for an MLFlow model registered on DagsHub.
+
+        Args:
+            open_project: Automatically open the Label Studio project in the browser
+            ignore_warning: Suppress the prompt-warning if you try to annotate too many datapoints at once.
+            fields_to_embed: list of meta-data columns that will show up in Label Studio UI.
+             if not specified all will be displayed.
+            fields_to_exclude: list of meta-data columns that will not show up in Label Studio UI
```
kbolashev, line 427: Leftovers from that other function to I suppose

dagshub/data_engine/model/query_result.py
kbolashev: Chat with Dean about a good name for this function. `get_predictions` is also very generic and might conflict with what I'm doing right now with the annotation conversions.
Probably needs a `from/with_mlflow_model` or smth like that

dagshub/data_engine/model/query_result.py
```diff
@@ -623,22 +694,63 @@ def visualize(self, visualizer: Literal["dagshub", "fiftyone"] = "dagshub", **kw
...
+            model_batch_size: function that converts to labelstudio format
         Returns:
             The URL of the created Label Studio workspace
         """
         send_analytics_event("Client_DataEngine_SentToAnnotation", repo=self.datasource.source.repoApi)
 
+        if all([param is not None for param in [model_name, model_version, model_repo, model_post_hook]]):
+            if use_remote_backend:
+                # send to dagshub backend
+                raise NotImplementedError("Coming soon!")
```
kbolashev, line 738: If this is not implemented then I don't see a point in having this is an argument. Delete it and we can come back to it once we need it

dagshub/data_engine/model/query_result.py
```diff
@@ -401,6 +404,74 @@ def download_binary_columns(
...
+            to the desired format
+            batch_size: (optional, default: 1) function that sets batch_size
+            metadata_column: (optional, default: 'prediction') write prediction results to metadatalogged in data engine.
+            If None, returns predictions.
+            return_predictions: (optional, default: True) returns predictions logged
+        Returns:
+            The URL of the created Label Studio workspace
+        """
+        if not metadata_column and not return_predictions:
+            raise ValueError("There is nothing to do. Either `metadata_column` or return_predictions` must be set!")
```
kbolashev, line 443:
```suggestion
            raise ValueError("Either `metadata_column` or `return_predictions` must be set.")
```

dagshub/data_engine/model/query_result.py
```diff
@@ -401,6 +404,74 @@ def download_binary_columns(
...
+
+        class TinyDL:
+            def __init__(self, dset, batch_size):
+                self.dset = dset
+                self.batch_size = batch_size
```
kbolashev, line 448: Better name for this pls, or a comment explaining what this is doing.
This reads very code golfy to me, especially with line 461 which ends up doing like 4 different logical jumps.

Is it a batcher?

dagshub/data_engine/model/query_result.py
```diff
@@ -401,6 +404,74 @@ def download_binary_columns(
...
+
+            def __iter__(self):
+                self.curr_idx = 0
+                return self
+
+            def __next__(self):
+                self.curr_idx += self.batch_size
+                return [self.dset[idx] for idx in range(self.curr_idx - self.batch_size, self.curr_idx)]
+
+        init(*repo.split("/")[::-1])
```
kbolashev, line 458: 1) You can get the repository from `self.datasource.source.repoApi`
2) What happens if init gets called multiple times? Honestly I never expected it to be used that way. Maybe we need to create an additional function that just sets up mlflow, so we don't get any weird side effects on this.

dagshub/data_engine/model/query_result.py
```diff
@@ -401,6 +404,74 @@ def download_binary_columns(
...
+                [result.path for result in self[idx * batch_size: (idx + 1) * batch_size]],
+            ):
+                predictions.append({remote_path: prediction})
+
+        if not metadata_column:
+            return predictions
+        with self.datasource.metadata_context() as ctx:
+            for remote_path in predictions:
+                ctx.update_metadata(remote_path, {metadata_column: predictions[remote_path]})
+        return predictions if return_predictions else True
```
kbolashev, line 475: Throw in a couple `log_message()`s here, since this is a very interactive function that also might have to wait a long time to wait for the predicting to finish.

If you want to go all the way, you can make it a progress bar with rich, metadata uploading has a working progress you can look at

dagshub/data_engine/model/query_result.py
```diff
@@ -401,6 +404,74 @@ def download_binary_columns(
...
+             if not specified all will be displayed.
+            fields_to_exclude: list of meta-data columns that will not show up in Label Studio UI
+            repo: repository to extract the model from
+            name: name of the model in the mlflow registry
+            version: (optional, default: 'latest') version of the model in the mlflow registry
+            pre_hook: (optional, default: identity function) function that runs before datapoint is sent to the model
+            post_hook: (optional, default: identity function) function that converts mlflow model output
+            to the desired format
+            batch_size: (optional, default: 1) function that sets batch_size
+            metadata_column: (optional, default: 'prediction') write prediction results to metadatalogged in data engine.
```
kbolashev: 1) We're not using `column`, we're trying to use `field` when talking about DE to prevent confusion with pandas dataframes.
2) Until reading the code this particular argument didn't make much sense to me. I think it'd be better as something like `log_to_field` or smth like that. Basically need to specify that this is actually an output and not an input.

kbolashev (changes requested):

dagshub/data_engine/model/query_result.py
kbolashev: Another note: what if I'm NOT using dagshub's mlflow, and a self-hosted one instead?
Maybe hide this behind two checks:
1) No active mlflow runs (if there is, means mlflow is set up already)
2) `mlflow.get_tracking_uri()` returns file location

dagshub/data_engine/model/query_result.py
jinensetpal: Yep. Line 461 checks if a batcher is required, and if not it just uses the base dataset, and iterates over it.

Added it as a comment for clarification!

dagshub/data_engine/model/query_result.py
jinensetpal: 1. The idea is that any model from any repo can be used - so I can use a model "off-the-shelf". I'll have it be optional, if they're just using the current repo.
2. The user will have to re-init with their repo before they can log a training run again. Right now, it's only setting mlflow anyway, but it needs to not persist. If they call init on their own repo multiple times, no change.
3. For the checks, I'll do the second one, but skip the first since models from different repos could be used, which would require re-initialization.

dagshub/data_engine/model/query_result.py
```diff
@@ -622,11 +697,44 @@ def visualize(self, visualizer: Literal["dagshub", "fiftyone"] = "dagshub", **kw
 
             return sess
 
+    def annotate_with_mlflow_model(
+        self,
```
kbolashev, line 701: Why does this function exist? Is it going to be doing something different later?
If they are doing the same thing basically, I think it will only lead to confusion as to which one needs to be used.

Also the docstring is wrong, don't see any Label Studio happening here

dagshub/data_engine/model/query_result.py
```diff
@@ -401,6 +404,78 @@ def download_binary_columns(
...
 
+    def predict_with_mlflow_model(
+        self,
+        repo: str,
+        name: str,
+        version: str = "latest",
+        pre_hook: Callable = lambda x: x,
+        post_hook: Callable = lambda x: x,
+        batch_size: int = 1,
+        log_to_field: str = None,
```
kbolashev, line 415:
```suggestion
        log_to_field: Optional[str] = None,
```

dagshub/data_engine/model/query_result.py
```diff
@@ -401,6 +404,78 @@ def download_binary_columns(
...
+        return_predictions: bool = True,
```
kbolashev, line 416: WDYT about getting rid of this argument completely and just always returning predictions?
It doesn't really cost anyone anything to return the resulting list in the end.

dagshub/data_engine/model/query_result.py
```diff
@@ -401,6 +404,78 @@ def download_binary_columns(
             num_proc=num_proc,
         )
 
+    def predict_with_mlflow_model(
+        self,
+        repo: str,
+        name: str,
+        version: str = "latest",
+        pre_hook: Callable = lambda x: x,
+        post_hook: Callable = lambda x: x,
```
kbolashev, line 413: Just so it's clear that you have one input and one output
```suggestion
        pre_hook: Callable[[Any], Any] = lambda x: x,
        post_hook: Callable[[Any], Any] = lambda x: x,
```

dagshub/data_engine/model/query_result.py
```diff
@@ -401,6 +404,78 @@ def download_binary_columns(
...
+        init(*repo.split("/")[::-1])
+        model = mlflow.pyfunc.load_model(f"models:/{name}/{version}")
+        init(*self.datasource.repo.split("/")[::-1])
+
+        dset = DagsHubDataset(self, tensorizers=[lambda x: x])
+
+        predictions = []
+        progress = get_rich_progress(rich.progress.MofNCompleteColumn())
+        task = progress.add_task("Running inference...", total=len(dset))
+        with progress:
```
kbolashev, line 461: ty for this!!!

dagshub/data_engine/model/query_result.py
```diff
@@ -401,6 +404,78 @@ def download_binary_columns(
...
+                return self
+
+            def __next__(self):
+                self.curr_idx += self.batch_size
+                return [self.dset[idx] for idx in range(self.curr_idx - self.batch_size, self.curr_idx)]
+
+        init(*repo.split("/")[::-1])
+        model = mlflow.pyfunc.load_model(f"models:/{name}/{version}")
+        init(*self.datasource.repo.split("/")[::-1])
+
```
kbolashev, line 455: Side effects of this:
1) Visual noise with the "Initialized repo blabla" from every init
2) If the user hasn't been using mlflow to track stuff before on this repo, now they do

Did you try going the MlflowClient route?
I saw that you tried to use `set_tracking_uri`, and honestly I think it'll be less noisy to use that function instead of init, if MlflowClient can't be used/too hard to use

dagshub/data_engine/model/query_result.py
```diff
@@ -401,6 +404,78 @@ def download_binary_columns(
...
+                ):
+                    predictions.append({remote_path: prediction})
+                progress.update(task, advance=batch_size, refresh=True)
+
+        if not log_to_field:
+            return predictions
+        with self.datasource.metadata_context() as ctx:
+            for remote_path in predictions:
+                ctx.update_metadata(remote_path, {log_to_field: predictions[remote_path]})
+        return predictions if return_predictions else True
```
kbolashev, line 477: If you delete the `return_predicitons` then this would be:

```suggestion
        if log_to_field:
            with self.datasource.metadata_context() as ctx:
                for remote_path in predictions:
                    ctx.update_metadata(remote_path, {log_to_field: predictions[remote_path]})
        return predictions
```

kbolashev (changes requested):
Almost there, thank you for taking time to do the progress bar!

dagshub/data_engine/model/query_result.py
jinensetpal: Agreed!

dagshub/data_engine/model/query_result.py
jinensetpal: I plan to add an additional parameter (the "Coming soon!" one), where instead of performing annotation locally it will create a remote LS model backend and perform inference using that.

Nonetheless, I think it's a good idea to have it because this use case might go under the radar, despite it as it currently stands just being a function wrapper.

jinensetpal: Thank you for the docstring catch, I've been slipping up with those. I'll fix that now!

dagshub/data_engine/model/query_result.py
```diff
@@ -401,6 +405,78 @@ def download_binary_columns(
...
+                return self
+
+            def __next__(self):
+                self.curr_idx += self.batch_size
+                return [self.dset[idx] for idx in range(self.curr_idx - self.batch_size, self.curr_idx)]
+
+        prev_uri = os.getenv("MLFLOW_TRACKING_URI", "")
+        os.environ["MLFLOW_TRACKING_URI"] = f"https://dagshub.com/{repo}.mlflow"
+        os.environ["MLFLOW_TRACKING_USERNAME"] = get_token()
+        os.environ["MLFLOW_TRACKING_PASSWORD"] = get_token()
```
kbolashev, line 452: OK, this still has one problem, which is that this only works with `dagshub.com`.

So what I suggest:
- Assume first that the repo with the model is on the same host as this repo. Get the host from `self.datasource.source.repoApi.host`.
- Use [multi_urljoin()](https://github.com/DagsHub/client/blob/dd3e1919281e5b2ebeaa247fe3ecc9ea615d2f14/dagshub/common/util.py#L11) to construct the TRACKING_URI without having to bother with remembering if there's a slash in the host or not: 
`multi_urljoin(host, f"{repo}.mlflow")`  
- Add a `host` argument to the functions, that allows you to override this behavior, if e.g. I'm on an on-prem installation and want to use a model from regular dagshub.com

P.S. also you can replace the `getenv` with `mlflow.get_tracking_uri()`, but this is a nitpick

kbolashev (changes requested):
Last little thing and we can merge it. Thank you!!!

jinensetpal:
Thank you for being so thorough!

kbolashev (approved):
Thank you, Jinen!! 🔥

kbolashev:
Closing as this is part of https://github.com/DagsHub/client/pull/513

PR #496, opened 2024-07-02
Enhancement/datetime-fucntions-annotations

sdafni:


dagshub/data_engine/model/datasource.py
```diff
@@ -370,7 +370,16 @@ def as_of(self, time: Union[float, datetime.datetime]) -> "Datasource":
...
+
+        A metadata of type datetime is always stored in DB as a UTC time, when a query is done on this field
+        there are 3 options:
+
+        - Metadata was saved with a timezone, in which case it will be used.
+
+        - Metadata was saved without a timezone, in which case UTC will be used.
+
+        - with_time_zone specified a time zone and it will override whatever is in the database.
         """
```
kbolashev, line 383: Can you put the link to the `with_time_zone` function on this? I had to dig into the source code to find where to specify that

dagshub/data_engine/dtypes.py
```diff
@@ -75,16 +75,29 @@ class DateTime(DagshubDataType):
...
+        # and the +05:30 offset
         datapoints = datasource.all()
         t = dateutil.parser.parse("2022-04-05T15:30:00.99999+05:30")
         datapoints[path][name] = t
         datapoints[path].save()
+
+        # or:
+        # send only a millisecond timestamp, without timezone (will be saved as utc)
+        datapoints[path][name] = int(dateutil.parser.parse("2022-04-05T15:30:00.99999+05:30").timestamp() * 1000)
     """
```
kbolashev, line 101: Get rid of the tz in the string here, I think it's confusing

kbolashev (approved):
Tiny changes, but otherwise good, and thank you for taking care of the docs <3

PR #501, opened 2024-07-18
Cache the current user profile

kbolashev:
This PR adds caching for the `UserAPI.get_user_from_token()` function.
This will lead to, e.g, caching of the current user, which will be useful for annotations later :)

### DIFF
--- a/dagshub/common/api/user.py
+++ b/dagshub/common/api/user.py
@@ -1,4 +1,4 @@
-from functools import cached_property
+from functools import cached_property, lru_cache
 from typing import Optional, Any, Union
 
 import dacite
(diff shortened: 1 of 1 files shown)


simonlsk (commented):
- [x] Let's discuss the security issue and decide.

guysmoilov:
One the one hand, I don't see a particular security issue since if the user has malicious code running in their process, then yes of course it can grab and steal anything. Do you imagine a more exotic threat scenario?

OTOH, why is this necessary? This caching helps with something frequent?

kbolashev:
> OTOH, why is this necessary? This caching helps with something frequent?

This will help with this PR: https://github.com/DagsHub/client/pull/503

In particular at the moment user clicks `dp.save()` with annotations, I have to get the current user to put into the LS task. Having the current user ID cached saves a round-trip every time they're saving annotations.

I initially threw `lru_cache` on the current user function, but it doesn't account for cases where the token was changed/they changed the user in the middle somehow.

simonlsk (approved):

PR #502, opened 2024-07-18
Add retry to the download function

kbolashev:
This PR adds retries on 5xx codes to the `download_files()` function.

This should mitigate problems occurring because of the server rate-limiting access to the /raw/ endpoint.

Example usecase that would cause the download to hit the rate limit:

```python
repo = RepoAPI("simon/baby-yoda-segmentation-dataset")
repo.download("images")
```

Before this PR user ends up with an incomplete directory. After this, it should eventually be downloaded (I have set the retries to max 5, so a long enough downtime would still fail)

### DIFF
--- a/dagshub/common/download.py
+++ b/dagshub/common/download.py
@@ -6,7 +6,8 @@ from functools import partial
 from pathlib import Path
 from typing import Tuple, Callable, Optional, List, Union, Dict
 
-from httpx import Auth
+from httpx import Auth, Response
+from tenacity import stop_after_attempt, wait_exponential, before_sleep_log, retry, retry_if_exception
 
 from dagshub.common import config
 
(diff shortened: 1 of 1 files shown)


dagshub/common/download.py
```diff
@@ -126,16 +127,29 @@ def download_url_to_bucket_path(url: str) -> Optional[Tuple[str, str, str]]:
     return groups["proto"], groups["bucket"], groups["path"]
 
 
+class DownloadError(Exception):
+    def __init__(self, response: Response):
+        self.response = response
+        super().__init__("Download failed with status code {response.status_code}")
```
simonlsk, line 133:
```suggestion
        super().__init__(f"Download failed with status code {response.status_code}")
```

simonlsk (approved):
I think you have a typo, but otherwise LGTM

dagshub/common/download.py
kbolashev: Thank you!

PR #505, opened 2024-07-18
Feature: Export a QueryResult with annotations as a YOLO dataset ready for training

kbolashev:
With this PR, It should be easier to train a YOLO model on a project with annotations:

```python
yaml_path = ds.all().download_as_yolo()

from ultralytics import YOLO
yolo = YOLO("yolov8n.pt")
yolo.train(data=yaml_path)
```

### DIFF
--- a/dagshub/data_engine/model/query_result.py
+++ b/dagshub/data_engine/model/query_result.py
@@ -1,13 +1,20 @@
 import inspect
 import logging
+from collections import Counter
 from concurrent.futures import as_completed, ThreadPoolExecutor
 from dataclasses import field, dataclass
 from os import PathLike
 from pathlib import Path
 from typing import TYPE_CHECKING, List, Dict, Any, Optional, Union, Tuple, Literal
+import os.path
 
 import dacite
+import dagshub_annotation_converter.converters.yolo
 import rich.progress
+from dagshub_annotation_converter.formats.yolo import YoloContext
+from dagshub_annotation_converter.formats.yolo.categories import Categories
+from dagshub_annotation_converter.formats.yolo.common import ir_mapping
+from dagshub_annotation_converter.ir.image import IRImageAnnotationBase
 
 from dagshub.common import config
 from dagshub.common.analytics import send_analytics_event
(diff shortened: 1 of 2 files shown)


dagshub/data_engine/model/query_result.py
```diff
@@ -482,6 +489,81 @@ def dp_url(dp: Datapoint):
...
+                ann_type = type(ann)
+                ann_types[ann_type] = ann_types.get(ann_type, 0) + 1
+            for cat in ann.categories.keys():
+                categories.get_or_create(cat)
+
+        if annotation_type is None:
+            annotation_types = [type(ann) for ann in annotations]
+            counter = Counter(annotation_types)
+            most_common = counter.most_common(1)
+            annotation_type = ir_mapping[most_common[0][0]]
```
simonlsk, line 542: What is `ir_mapping` ? 
Why does `type(<something>)` return an array of arrays? 
Both things are not verbose enough IMO, both requires reading the further and I can't deduce what those functions do from reading this code.

dagshub/data_engine/model/query_result.py
```diff
@@ -482,6 +489,81 @@ def dp_url(dp: Datapoint):
...
+
+        # Add the source prefix to all annotations
+        for ann in annotations:
+            ann.filename = os.path.join(self.datasource.source.source_prefix, ann.filename)
+
+        # If there's no folder "images" in the datasource, prepend it to the path
+        if not any("images/" in ann.filename for ann in annotations):
+            for ann in annotations:
+                ann.filename = os.path.join("images", ann.filename)
+            image_download_path = Path(download_dir) / "images"
```
simonlsk, line 555: So for example images will be downloaded to `$pwd/dagshub_export/data/images`

dagshub/data_engine/model/query_result.py
```diff
@@ -482,6 +489,81 @@ def dp_url(dp: Datapoint):
...
+
+        context = YoloContext(annotation_type=annotation_type, categories=categories)
```
simonlsk, line 557: You should probably allow pass through of categories as well, I guess.

dagshub/data_engine/model/query_result.py
```diff
@@ -482,6 +489,81 @@ def dp_url(dp: Datapoint):
...
+        context.path = image_download_path
+
+        log_message("Downloading image files...")
+        self.download_files(image_download_path)
```
simonlsk, line 561: Can I not download the actual images if I have them?

setup.py
```diff
@@ -42,6 +42,10 @@ def get_version(rel_path: str) -> str:
     "python-dateutil",
     "tenacity~=8.2.3",
     "boto3",
+    # FIXME: GO BACK TO REGULAR IMPORT AFTER THIS IS MERGED
+    "dagshub-annotation-converter @ "
+    + "git+https://github.com/DagsHub/"
+    + "dagshub-annotation-converter@refactor/static-annotations#egg=dagshub-annotation-converter",
```
simonlsk, line 48: - [ ] Not forget

simonlsk (changes requested):
Pretty straight forward, looks good overall.
I have a few suggestions and questions.

dagshub/data_engine/model/query_result.py
kbolashev: If you have them at the exact path they will be downloaded to, then the downloader will skip them.
I don't expect users to guess exactly where it will get downloaded to though to be honest

dagshub/data_engine/model/query_result.py
kbolashev: Only if there's no `images` in the paths. I think if you have your datasource at `/data/images`, that will actually happen already, since I'm not exporting with the whole repository path.
So a dp `1.png` will just get downloaded to `1.png`, and I want it to be in `images/1.png`.

I need to have the `images` in the paths for the whole `images <-> labels` switch to work, so that's why I add it here.

dagshub/data_engine/model/query_result.py
kbolashev: Is there a usecase for it? because it's not like YOLO can annotate stuff that wasn't in the training data.

Besides allowing users to annotate their own stuff will also need some safety checking, what if there are missing categories?

dagshub/data_engine/model/query_result.py
kbolashev: ir_mapping is a mapping from the intermidary classes to the YOLO types.
So if it's an `IRImageSegmentationAnnotation`, it will be `segmentation` for example.

will it be better if i rename it to `ir_to_yolo_type_mapping`?

PR #506, opened 2024-07-21
Bug: Filter the metadata fields based on the introspection

kbolashev:
New timeZone field is not yet deployed in production, and because of that the queries are failing on the current version of the client.

I added introspection-based filtering that removes the timeZone field if it's not included in the GQL schema.
This is not a big deal for the consuming code, because the consuming code uses `metadata.get("timeZone")` later to get the time zone, falling back to +00:00, so should be ok.


EvgeniLeonti (approved):
LGTM

PR #509, opened 2024-07-30
Enhancement: Add explanation of what install_hooks() is doing

kbolashev:
Running `install_hooks` will now show an explanation to the user of what's happening, so as to not feel too magical.
Example:
```
Repository "Dean/COCO_1K" is now hooked at path "/Users/kirillbolashev/Job/GitRepos/client-test-suite/manual_tests/annotations/coco_1k".
Any calls to Python file access function like open() and listdir() inside of this directory will include results from the repository.

```


simonlsk (approved):
LGTM,
Thanks

PR #511, opened 2024-08-01
Datasource: Handle pandas timestamp in df upload

kbolashev:
Currently uploading a dataframe with pandas' Timestamp objects fails.
This PR fixes it, converting them to datetimes.


guysmoilov (approved):
OK.
I think the more elegant systematic solution might be to check the superclasses but maybe that's too aggressive

PR #512, opened 2024-08-01
Handle datetimes correctly for dataframe + multivalue

kbolashev:
The backend expects datetimes to be sent as timestamp + additionally a timeZone field in the metadata, if there's a timezone.

The dataframe code wasn't handling that right.
This bug fixes it.

Made sure to test an upload locally from a dataframe, to be sure that it's working e2e now. Got the same data I uploaded back


guysmoilov (approved):
Oh my, that is indeed not DRY.
But, if it works...

PR #517, opened 2024-08-14
enhancement / remote metadata upload 

EvgeniLeonti:
Main changes:
1. extend existing `upload_metadata_from_dataframe` to include `remote` flag
2. new method `upload_metadata_from_file` which will allow uploading metadata from a file + include `remote` flag
3. when remote is false - fallback to existing behavior
4. when remote is true - call our backend with `import/metadata` endpoint which will use python bgproc that also using the client with but with remote = false (meaning that dataengine api calls will be done in remote environment)

### DIFF
--- a/dagshub/common/api/repo.py
+++ b/dagshub/common/api/repo.py
@@ -1,7 +1,7 @@
 import logging
 from os import PathLike
 from pathlib import Path, PurePosixPath
-
+import mimetypes
 import rich.progress
 
 from dagshub.common.api.responses import (
@@ -388,6 +388,30 @@ class RepoAPI:
         download_files(file_tuples, skip_if_exists=not redownload)
         log_message(f"Downloaded {len(files)} file(s) to {local_path.resolve()}")
 
+    def import_metadata_from_file(self, datasource_name: str, file_path: Union[str, PathLike], path_column: str):
+        files = {
+            "file": (
+                Path(file_path).name,
+                open(file_path, "rb"),
+                mimetypes.guess_type(file_path)[0] or "application/octet-stream",
+            ),
+        }
+
+        data = {
+            "datasource_name": datasource_name,
+            "path_column": path_column,
+        }
+
+        res = self._http_request("POST", self.repo_import_metadata_from_file_url(), data=data, files=files)
+        if res.status_code == 404:
+            raise PathNotFoundError(f"Datasource {datasource_name} not found")
+        elif res.status_code >= 400:
+            error_msg = f"Got status code {res.status_code} when importing metadata to datasource {datasource_name}"
+            logger.error(error_msg)
+            logger.debug(res.content)
+            raise RuntimeError(error_msg)
+        return res.content
+
     @staticmethod
     def _sanitize_storage_path(path: Union[str, PathLike]) -> Tuple[str, bool]:
         """
@@ -562,6 +586,9 @@ class RepoAPI:
         :meta private:
         """
         return multi_urljoin(self.host, "api/v1/repo-buckets/s3", self.owner)
+
+    def repo_import_metadata_from_file_url(self) -> str:
+        return multi_urljoin(self.data_engine_url, "import/metadata")
 
     @staticmethod
     def parse_repo(repo: str) -> Tuple[str, str]:
--- a/dagshub/common/api/repo.py
+++ b/dagshub/common/api/repo.py
@@ -1,7 +1,6 @@
 import logging
 from os import PathLike
 from pathlib import Path, PurePosixPath
-import mimetypes
 import rich.progress
 
 from dagshub.common.api.responses import (
@@ -388,30 +387,6 @@ class RepoAPI:
         download_files(file_tuples, skip_if_exists=not redownload)
         log_message(f"Downloaded {len(files)} file(s) to {local_path.resolve()}")
 
-    def import_metadata_from_file(self, datasource_name: str, file_path: Union[str, PathLike], path_column: str):
-        files = {
-            "file": (
-                Path(file_path).name,
-                open(file_path, "rb"),
-                mimetypes.guess_type(file_path)[0] or "application/octet-stream",
-            ),
-        }
-
-        data = {
-            "datasource_name": datasource_name,
-            "path_column": path_column,
-        }
-
-        res = self._http_request("POST", self.repo_import_metadata_from_file_url(), data=data, files=files)
-        if res.status_code == 404:
-            raise PathNotFoundError(f"Datasource {datasource_name} not found")
-        elif res.status_code >= 400:
-            error_msg = f"Got status code {res.status_code} when importing metadata to datasource {datasource_name}"
-            logger.error(error_msg)
-            logger.debug(res.content)
-            raise RuntimeError(error_msg)
-        return res.content
-
     @staticmethod
     def _sanitize_storage_path(path: Union[str, PathLike]) -> Tuple[str, bool]:
         """
@@ -586,9 +561,6 @@ class RepoAPI:
         :meta private:
         """
         return multi_urljoin(self.host, "api/v1/repo-buckets/s3", self.owner)
-
-    def repo_import_metadata_from_file_url(self) -> str:
-        return multi_urljoin(self.data_engine_url, "import/metadata")
 
     @staticmethod
     def parse_repo(repo: str) -> Tuple[str, str]:
--- a/dagshub/data_engine/model/datasource.py
+++ b/dagshub/data_engine/model/datasource.py
@@ -2,6 +2,7 @@ import base64
 import datetime
 import json
 import logging
+import tempfile
 import math
 import os.path
 import threading
@@ -567,19 +568,13 @@ class Datasource:
     def _remote_upload_metadata_from_dataframe(
         self, df: "pandas.DataFrame", path_column: Optional[Union[str, int]] = None
     ):
-        random_file_name = str(uuid.uuid4())
-        file_path = f"{random_file_name}.parquet"
-        df.to_parquet(file_path, index=False)
         datasource_name = self.source.name
 
-        task_result = None
-        try:
-            task_result = self.source.repoApi.import_metadata_from_file(datasource_name, file_path, path_column)
-        except:  # noqa # todo better exception handling when we'll have feedback from the backend
-            pass
-        finally:
-            os.remove(file_path)
-        return task_result
+        with tempfile.NamedTemporaryFile(suffix=".parquet", delete=True) as tmp:
+            file_path = tmp.name
+            df.to_parquet(file_path, index=False)
+
+            self.source.repoApi.import_metadata_from_file(datasource_name, file_path, path_column)
 
     def _get_multivalue_fields(self) -> Set[str]:
         res = set()
--- a/dagshub/data_engine/model/datasource.py
+++ b/dagshub/data_engine/model/datasource.py
@@ -514,7 +514,9 @@ class Datasource:
 
         return func()
 
-    def upload_metadata_from_file(self, file_path, path_column: Optional[Union[str, int]] = None, remote: bool = False):
+    def upload_metadata_from_file(
+        self, file_path, path_column: Optional[Union[str, int]] = None, ingest_on_server: bool = False
+    ):
         """
         Upload metadata from a file.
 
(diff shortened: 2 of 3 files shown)


dagshub/data_engine/model/datasource.py
```diff
@@ -524,10 +551,35 @@ def upload_metadata_from_dataframe(self, df: "pandas.DataFrame", path_column: Op
...
+        datasource_name = self.source.name
+
+        task_result = None
+        try:
+            task_result = self.source.repoApi.import_metadata_from_file(datasource_name, file_path, path_column)
+        except:  # noqa # todo better exception handling when we'll have feedback from the backend
+            pass
+        finally:
+            os.remove(file_path)
+        return task_result
```
kbolashev, line 582: Use [tempfile](https://docs.python.org/3/library/tempfile.html#examples) here (TemporaryDirectory in particular), will save you from doing the os.remove, the lib will do it for you.

Also you don't need to have an `except` here, python is fine with just this:
```
try:
    failing_code()
finally:
    cleanup()
```

Cleanup will always run, and the exception will still be thrown

dagshub/data_engine/model/datasource.py
```diff
@@ -513,7 +513,34 @@ def func():
...
+        """
+        Upload metadata from a file.
+
+        Args:
+            file_path: Path to the file with metadata. Allowed formats are CSV, Parquet, ZIP, GZ.
+            path_column: Column with the datapoints' paths. Can either be the name of the column, or its index.
+                If not specified, the first column is used.
+            remote: Whether to upload the metadata via remote or local upload. Default is False.
+                Remote upload means that the data engine API calls will be done from remote machine,
+                rather than from local machine which runs this client.
```
kbolashev, line 526: [blocker]
This needs another name, remote is very unintuitive. Check the comments in the design doc
I don't have a good alternative suggestion though to be honest.

dagshub/data_engine/model/datasource.py
```diff
@@ -1391,6 +1443,21 @@ def _test_not_comparing_other_ds(other):
...
+        if file_path.lower().endswith(".csv"):
+            df = pd.read_csv(file_path)
+        elif file_path.lower().endswith(".parquet"):
+            df = pd.read_parquet(file_path)
+        elif file_path.lower().endswith(".zip"):
+            df = pd.read_csv(file_path, compression="zip")
+        elif file_path.lower().endswith(".gz"):
+            df = pd.read_csv(file_path, compression="gzip")
+        else:
+            raise RuntimeError(f"Unsupported file format: {file_path}")
```
kbolashev, line 1458: Let the user know what they can import + a whole path is more than just the format, so the error could be more precise.
```suggestion
        else:
            raise RuntimeError(f"File '{file_path}' needs to be a .csv/.parquet or a compressed .zip/.gz to be imported")
```

btw, I would also look into how the compression flags work, because I expect users throwing a 10-file archives at this function and it failing in pandas. Can you check what pandas throws when you do that?

dagshub/data_engine/model/datasource.py
```diff
@@ -513,7 +513,34 @@ def func():
...
+        """
+        self.source.get_from_dagshub()
+        send_analytics_event("Client_DataEngine_addEnrichmentsWithFile", repo=self.source.repoApi)
+
+        if remote:
+            try:
+                datasource_name = self.source.name
+                self.source.repoApi.import_metadata_from_file(datasource_name, file_path, path_column)
+            except:  # noqa # todo better exception handling when we'll have feedback from the backend
+                pass
```
kbolashev, line 536: I'm against catch-all exception swallowing. I suggest just letting it run as is.

dagshub/data_engine/model/datasource.py
```diff
@@ -513,7 +513,34 @@ def func():
...
+
+        Args:
+            file_path: Path to the file with metadata. Allowed formats are CSV, Parquet, ZIP, GZ.
+            path_column: Column with the datapoints' paths. Can either be the name of the column, or its index.
+                If not specified, the first column is used.
+            remote: Whether to upload the metadata via remote or local upload. Default is False.
+                Remote upload means that the data engine API calls will be done from remote machine,
+                rather than from local machine which runs this client.
+        """
+        self.source.get_from_dagshub()
```
kbolashev, line 528: Did you have to add this due to something that showed up during the testing?
In general this function should only be called _after_ uploading metadata (synchronously) to make sure that the field schema in the client is synchronized with the backend (for e.g. making sure the users filter on existing fields).

For this particular call this seems redundant

dagshub/data_engine/model/datasource.py
```diff
@@ -524,10 +551,35 @@ def upload_metadata_from_dataframe(self, df: "pandas.DataFrame", path_column: Op
                 DataFrame with metadata
             path_column: Column with the datapoints' paths. Can either be the name of the column, or its index.
                 If not specified, the first column is used.
+            remote: Whether to upload the metadata via remote or local upload. Default is False.
+                Remote upload means that the data engine API calls will be done from remote machine,
+                rather than from local machine which runs this client.
```
kbolashev, line 556: Same comment re: naming.

kbolashev (changes requested):
Can you add ability to use files from the repository?
I feel like this will push it from "cool" to "really useful" territory.

dagshub/data_engine/model/datasource.py
kbolashev: Example: https://github.com/DagsHub/client/blob/34a0c7a44db462e9881125f23c1a3ea93ad5b7ad/dagshub/notebook.py#L95

dagshub/data_engine/model/datasource.py
EvgeniLeonti: In such cases it would throw:
```
                    raise ValueError(
                        "Multiple files found in ZIP file. "
                        f"Only one file per ZIP: {zip_names}"
                    )
```

which evaluates to:
```
ValueError: Multiple files found in ZIP file. Only one file per ZIP: ['squirrel_detection.gz', '__MACOSX/._squirrel_detection.gz', 'squirrel_detection.parquet', '__MACOSX/._squirrel_detection.parquet']
```

do you think its better to catch this error and return custom error message from the client?

dagshub/common/api/repo.py
```diff
@@ -388,6 +388,30 @@ def download(
...
+
+        res = self._http_request("POST", self.repo_import_metadata_from_file_url(), data=data, files=files)
+        if res.status_code == 404:
+            raise PathNotFoundError(f"Datasource {datasource_name} not found")
+        elif res.status_code >= 400:
+            error_msg = f"Got status code {res.status_code} when importing metadata to datasource {datasource_name}"
+            logger.error(error_msg)
+            logger.debug(res.content)
+            raise RuntimeError(error_msg)
+        return res.content
```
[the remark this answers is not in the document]
EvgeniLeonti: How would you do that?

Do you mean call existing `RepoAPI.import_metadata_from_file` from `DatasourceState` or implement the request itself inside `DatasourceState`?

dagshub/data_engine/model/datasource.py
EvgeniLeonti: Few suggestions, let me know what makes sense the most:
1. remote_upload
2. remote_execution
3. backgroud_processing

dagshub/data_engine/model/datasource.py
EvgeniLeonti: Will do once we decide about a name.

dagshub/common/api/repo.py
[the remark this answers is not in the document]
kbolashev: Second one, just cut-paste the function into there.
I know that the name is `DatasourceState`, but I kind of made it to be a place where you would have most of the lower-level interactions with the data engine backend. Meanwhile the datasource is what the user interacts with.

dagshub/data_engine/model/datasource.py
kbolashev: I like Guy's suggestion of `ingest_on_server`, let me know what you think about it.
Maybe it can be shortened to just `serverside`.

Also alternative wording for the docstring, I'm rewording that because the client user doesn't really have a concept of "background processing", or what is a "remote machine":

```
ingest_on_server: Set to ``True`` to process the metadata asynchronously. The file will be sent to our server and ingested into the datasource there.
```

dagshub/common/api/repo.py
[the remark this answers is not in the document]
EvgeniLeonti: Copy-paste won't work out of the box because `DatasourceState` doesn't have `self._http_request`. It uses the REST API via `self.repoApi` and GQL API via `self.client`.

How would you tackle that?
Adding or re-using`self._http_request` seems just wrong because we have `self.repoApi` for that.
Creating a wrapper function in `DatasourceState` seems redundant overhead, but at least it would keep using the convention that REST API calls made via `self.repoApi`.

WDYT?

dagshub/data_engine/model/datasource.py
kbolashev: No need to catch it, imo it's fine, looks pretty self explanatory!

dagshub/data_engine/model/datasource.py
EvgeniLeonti: `ingest_on_server` sounds pretty good. Agree about the "remote machine" and "background processing" terminology.

EvgeniLeonti:
> Can you add ability to use files from the repository? I feel like this will push it from "cool" to "really useful" territory.

Sure.

What will be more clear?
1. rename `import_metadata_from_file` to `import_metadata_from_local_file`; introduce `import_metadata_from_reoi`
2. re-use `import_metadata_from_file` by adding arg like `file_location = local / repo`

dagshub/common/api/repo.py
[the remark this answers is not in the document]
kbolashev: `DatasourceState` still has `repoAPI` in it, so it can call `state.repoApi._http_request`.
It's not really a good practice to use underscored functions, but it's not like they are actually private, you can still call them.

My problem with having this function in RepoAPI is that RepoAPI might be useful for a user, so we're trying to keep the functions there useful for a regular user and have them make sense. This function is too tightly bound to Data Engine to really be put in RepoAPI imo, since it expects to have a datasource specified in the first place.

Keep in mind that this function will basically show up here, after you write a docstring for it: https://dagshub.com/docs/client/reference/repo_api.html and it will also show up in autocompletion.

IMO putting this in `DatasourceState` away from public functions, and using a private `repoAPI._http_request` is the best compromise

kbolashev:
> > Can you add ability to use files from the repository? I feel like this will push it from "cool" to "really useful" territory.
> 
> Sure.
> 
> What will be more clear?
> 
>     1. rename `import_metadata_from_file` to `import_metadata_from_local_file`; introduce new `import_metadata_from_repo`
> 
>     2. re-use `import_metadata_from_file` by adding arg like `file_location = local / repo`

I would want it to be "automagic" and try to first look for the file locally, failing that, try to find it in repo, and failing that throw.
I've already implemented a similar thing in the annotation import PR: https://github.com/DagsHub/client/blob/038d864ae07ef28f6aa578d6b4e357c967eeaa42/dagshub/data_engine/annotation/importer.py#L109

However, that still can use some improvements, since the function doesn't handle buckets. Maybe we can deploy this right now, and I can work on implementing a generic `RepoAPI.has_file()` and `RepoAPI.get_file()` that handles repo buckets correctly, and then we will reuse that here.

dagshub/data_engine/model/datasource.py
```diff
@@ -14,7 +15,7 @@
 from pathlib import Path
 from typing import Any, Dict, List, Optional, TYPE_CHECKING, Union, Set, ContextManager, Tuple, Literal
 
-
+import pandas as pd
```
kbolashev, line 18: `pandas` is already lazy loaded at line 67, so adding an eager import is redundant.
Get rid of this and swap the calls in `_convert_file_to_df` to use `pandas` instead of `pd`

kbolashev (changes requested):
One last small thing I noticed, anything else I don't really see as blockers for MVP

EvgeniLeonti:
> > > Can you add ability to use files from the repository? I feel like this will push it from "cool" to "really useful" territory.
> > 
> > 
> > Sure.
> > What will be more clear?
> > ```
> > 1. rename `import_metadata_from_file` to `import_metadata_from_local_file`; introduce new `import_metadata_from_repo`
> > 
> > 2. re-use `import_metadata_from_file` by adding arg like `file_location = local / repo`
> > ```
> 
> I would want it to be "automagic" and try to first look for the file locally, failing that, try to find it in repo, and failing that throw. But also with a way to specify the file location explicitly, so kind of the second case I've already implemented a similar thing in the annotation import PR:
> 
> https://github.com/DagsHub/client/blob/038d864ae07ef28f6aa578d6b4e357c967eeaa42/dagshub/data_engine/annotation/importer.py#L109
> 
> However, that still can use some improvements, since the function doesn't handle buckets. Maybe we can deploy this right now, and later we can work on implementing a generic `RepoAPI.has_file()` and `RepoAPI.get_file()` that handles repo buckets correctly, and then we will reuse that here.

Implemented that similar to what you showed in annotation import, with a TODO to reuse the generic solution once improved.

kbolashev (commented):
I didn't merge the PR with that function yet 😢 
Blocked by annotation converter review for now
Can you copy it in here maybe?

EvgeniLeonti:
I copied it as `_determine_load_location` in `datasource_state` with small modification to fit the context.
Do you mean another kind of copy?

kbolashev:
Ah, no, sorry, I just completely glossed by it. My apologies

kbolashev (approved):

PR #522, opened 2024-09-03
Infra: Unpin semver dependencies

kbolashev:
Some of the version pinning was way too agressive, conflicting with Airflow.

This is a WIP, for now I just unpinned every dependency and made it from `~=` to `>=`, will see if it passes the tests and our backend test suite.


kbolashev:
The backend suite seems ok, merging

PR #523, opened 2024-09-05
feature/autolog datasource id

simonlsk:
## Spec
* Change the storing format of datasource and dataset to allow multi dataset / datasource run.


kbolashev (approved):

simonlsk:
Not relevant anymore, the details should belong to the QueryResult instead.

PR #529, opened 2024-09-23
Enhancement: Use timestamp returned from the server in QueryResult

kbolashev:
Scope of this PR:
- Fetch the `queryDataTime` field from the GraphQL of the query, put the value in `QueryResult.query_data_time` (convert to datetime, assume `queryDataTime` is a UTC timestamp)
- Add an `as_of` argument to `Datasource.log_to_mlflow()`, that specifies the query timestamp
- If autologging to MLflow is on (there is an active MLflow run), then `QueryResult.query_data_time` is passed to the MLflow logging function.
- Add a `DatasourceSerializedState.timed_link` field, which has different semantics from `DatasourceSerializedState.link`:
  - If there is a global `as_of` already set up on the query, then it is the same as `link`
  - Otherwise the `as_of` from the Query is used there (or datetime.now if it was logged by the user with no as_of)

I decided to go the way of adding a new argument to a function so as to not move the whole `log_to_mlflow` function from `Datasource` to `QueryResult`. Considering that we mostly not expect `log_to_mlflow` to be called by users, and instead they should utilize autologging, this is an OK compromise.

The only UX friction I envision is using `Datasource.log_to_mlflow()` will now have an additional optional argument, and if the user is using that to save their datasources, they might miss that the time can also be saved there. If you consider that a problem, I can go back to the drawing board and migrate the function to `QueryResult.log_to_mlflow`

### DIFF
--- a/dagshub/data_engine/model/datasource.py
+++ b/dagshub/data_engine/model/datasource.py
@@ -844,7 +844,7 @@ class Datasource:
             return
         source_name = self.source.name
 
-        now_time = datetime.datetime.now().strftime("%Y-%m-%dT%H-%M-%S")  # Not ISO format to make it a valid filename
+        now_time = qr.query_data_time.strftime("%Y-%m-%dT%H-%M-%S")  # Not ISO format to make it a valid filename
         uuid_chunk = str(uuid.uuid4())[-4:]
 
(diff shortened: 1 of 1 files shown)


guysmoilov (approved):
As you said, it seems messy but functional

PR #531, opened 2024-09-25
Add a fallback on annotation validation

kbolashev:
Having a malformed annotation on a datapoint (or an annotation that failed to parse, not even malformed) right now leads to the whole loading process failing.

This PR is a hotfix to prevent that.
If an annotation fails to parse, it gets shown to the user in a warning message, and I warn them that they won't be able to interact with them, for example:

```
Accessing as kirill
Downloading annotation fields ['annotation']...
Warning: The following datapoints had invalid annotations, any annotation-related operations will not work on these:
Field "annotation" in datapoints:
        train2017/000000000009.jpg
```

I'm not handling these bad annotations in other downstream functions for now properly, if we do want to make it have a nice UX, would probably need to have some kind of `BrokenAnnotation` class that will shim the regular annotation functions and display a warning.
Not sure if it's worth it to work on this right now

### DIFF
--- a/dagshub/data_engine/model/query_result.py
+++ b/dagshub/data_engine/model/query_result.py
@@ -1,6 +1,6 @@
 import inspect
 import logging
-from collections import Counter
+from collections import Counter, defaultdict
 from concurrent.futures import as_completed, ThreadPoolExecutor
 from dataclasses import field, dataclass
 from os import PathLike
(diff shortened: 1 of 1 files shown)


EvgeniLeonti (approved):
LGTM

Make sure formatting is ok:
```
dagshub/data_engine/model/query_result.py#L446
Too many blank lines (2) (E303)
```

kbolashev:
Good catch, I think my autoformatter broke 🥲

PR #535, opened 2024-10-05
Fix but in 'LS backend with MLflow model' flow

idonov8:
I tried running and got:
```
File /usr/local/lib/python3.11/site-packages/tenacity/__init__.py:382, in Retrying.__call__(self, fn, *args, **kwargs)
    380 if isinstance(do, DoAttempt):
    381     try:
--> 382         result = fn(*args, **kwargs)
    383     except BaseException:  # noqa: B902
    384         retry_state.set_exception(sys.exc_info())  # type: ignore[arg-type]

File /usr/local/lib/python3.11/site-packages/dagshub/data_engine/model/datasource.py:1140, in Datasource.add_annotation_model(self, repo, name, version, post_hook, pre_hook, port, project_name, ngrok_authtoken)
   1138     if not self.ngrok_listener:
   1139         self.ngrok_listener = ngrok.forward(port, authtoken=ngrok_authtoken)
-> 1140     endpoint = self.ngrok_listener.url()
   1141 else:
   1142     endpoint = f"{LS_ORCHESTRATOR_URL}:{port}/"

AttributeError: '_asyncio.Task' object has no attribute 'url'
```

This PR solves this issue without changing any interface.

### DIFF
--- a/dagshub/data_engine/model/datasource.py
+++ b/dagshub/data_engine/model/datasource.py
@@ -57,10 +57,7 @@ from dagshub.data_engine.model.metadata_field_builder import MetadataFieldBuilde
 from dagshub.data_engine.model.query import QueryFilterTree
 from dagshub.data_engine.model.schema_util import metadataTypeLookup, metadataTypeLookupReverse
 from dagshub.data_engine.model.datasource_state import DatasourceState
-# 👇 Required for running adding of LS backend from a notebook
-import asyncio
-import nest_asyncio
-nest_asyncio.apply()
+
 if TYPE_CHECKING:
     from dagshub.data_engine.model.query_result import QueryResult
     import fiftyone as fo
@@ -1039,7 +1036,7 @@ class Datasource:
         return self.source.metadata_fields
 
     @retry(retry=retry_if_exception_type(LSInitializingError), wait=wait_fixed(3), stop=stop_after_attempt(5))
-    def add_annotation_model_from_config(self, config, project_name, ngrok_authtoken, port=9090):
+    async def add_annotation_model_from_config(self, config, project_name, ngrok_authtoken, port=9090):
         """
         Initialize a LS backend for ML annotation using a preset configuration.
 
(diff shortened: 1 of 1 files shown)


dagshub/data_engine/model/datasource.py
```diff
@@ -57,7 +57,10 @@
 from dagshub.data_engine.model.query import QueryFilterTree
 from dagshub.data_engine.model.schema_util import metadataTypeLookup, metadataTypeLookupReverse
 from dagshub.data_engine.model.datasource_state import DatasourceState
-
+# 👇 Required for running adding of LS backend from a notebook
+import asyncio
+import nest_asyncio
+nest_asyncio.apply()
```
kbolashev, line 63: 1) This new package hasn't been added to the requirements
2) The package is archived as of May this year: https://github.com/erdewit/nest_asyncio
I am not super comfortable adding a dead package like that, considering that it's doing things to the python internals that are possible to change on a new Python release. Is there another way to do this? What happens if you spin off a regular thread?

kbolashev (changes requested):

kbolashev:
Wait, I see the exception is returning a Task, seems like a problem with the function

kbolashev:
Found this: https://www.github.com/ngrok/ngrok-python/issues/125
Seems like this happened because you called it from inside of an async function already.

Messed up that they are doing this. I guess the only solution I see that doesn't involve writing a lot of code to catch this would be to mark this function async and leave handling async up to the user.

dagshub/data_engine/model/datasource.py
kbolashev: Also just as a general note, putting this at the top of the file means it will be applied to EVERYONE who ever uses the client.
This is a pretty dangerous and intrusive change, so if we do need to resort to this in the end, it should be localized to the ngrok forwarding function only.

kbolashev:
Also very happy to see you back, Ido! 🥳

idonov8:
> I guess the only solution I see that doesn't involve writing a lot of code to catch this would be to mark this function async and leave handling async up to the user.

I don't mind it, but it means that the user will need to run `await ds.add_annotation_model_from_config(...)`
That'll require some changes to the docs as well. Is that ok?

kbolashev:
I don't mind it personally. It's a better compromise than basically leaving it completely broken IF the user is using async.

I'll ask Dean, maybe he has a stronger opinion about it

kbolashev (approved):
I checked that Tenacity can retry async functions, so seems like we should be ok on this.

You'll probably have to change some (hopefully not a lot) of the code that actually uses it :/

PR #545, opened 2024-11-04
Enhancement: Override config.host in dagshub.init

kbolashev:
Allows user to do:

```python
import dagshub
import mlflow

dagshub.init(repo_owner="user", repo_name="repo", host="https://some.other.dagshub")
with mlflow.start_run():
    mlflow.log_metric("aaa", 5)
```

The init will override the `config.host`, so all of the calls to dagshub will call the specified host instead.

### DIFF
--- a/dagshub/common/config.py
+++ b/dagshub/common/config.py
@@ -61,3 +61,12 @@ if download_threads > DEFAULT_DOWNLOAD_THREADS:
         f"Number of download threads was set to {download_threads}. "
         f"We recommend lowering the value if you get met with rate limits"
     )
+
+
+def set_host(new_host: str):
+    _parsed_host = urlparse(new_host)
+    _hostname = _parsed_host.hostname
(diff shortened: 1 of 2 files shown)


dagshub/common/config.py
```diff
@@ -61,3 +61,12 @@
...
     )
+
+
+def set_host(new_host: str):
+    _parsed_host = urlparse(new_host)
+    _hostname = _parsed_host.hostname
+    _host = _parsed_host.geturl().rstrip("/")
+
+    global hostname, host, parsed_host
+    hostname, host, parsed_host = _hostname, _host, _parsed_host
```
guysmoilov, line 72: Shouldn't this logic be reused somehow in the piece of code where you read DAGSHUB_CLIENT_HOST ?

kbolashev: Honestly not a huge fan of how it looks like with the reuse, but sure.

PR #547, opened 2024-11-07
Expose rclone_init in the storage docs

kbolashev:
Expose the `rclone_init` function that generates the rclone config in the documentation of the storage bucket + add it as a top-level import for `dagshub.storage`


guysmoilov (approved):
Nice. But, can you maybe add to the https://dagshub.com/docs/client/reference/repo_bucket.html#dagshub.storage.mount doc a sentence explaining that it uses rclone for mounting the repo bucket, and will offer installation instructions if it's missing?

PR #557, opened 2024-12-05
Enhancement: Make the visualization URL shorter if there's no query

kbolashev:
This PR shortens the encoded JSON that gets sent for visualization.
Before this PR an empty query will result in encoding a JSON like:
```json
{"query": null}
```

After this PR the null values will be omitted, so the result would be just `{}`


simonlsk (approved):
LGTM, thanks!

PR #560, opened 2024-12-08
Enhancement: Add clear_token_cache function

kbolashev:
Add a `dagshub.auth.clear_token_cache` function that removes the token file after prompting a user.

Also added a troubleshooting page that references this function in cases where a user can't connect to their repository.


EvgeniLeonti (commented):
LGTM!

I would change the default to be `No` to avoid cases of accidentally pressing enter

PR #561, opened 2024-12-08
Bug: Loading a model from the root of a bucket has to have a trailing slash in the path

kbolashev:
Fixes the usecase where:

- A model is stored in the root of a bucket like `bucket/some/prefix`
- The model is trying to be loaded with the following code:
```python
# This works
# bucket_name='bucket/some/prefix/'
# This fails without a trailing slash
bucket_name='bucket/some/prefix'

get_model_path("user/repo", path=bucket_name)
```

Switching for the bucket name with the slash works.

Also added a test case for this bug.

Most of the lines changed are reformats

### DIFF
--- a/dagshub/models/model_locator.py
+++ b/dagshub/models/model_locator.py
@@ -86,11 +86,7 @@ class ModelLocator:
     def download_destination(self) -> Path:
         if self._download_dest is not None:
             return Path(self._download_dest)
-        return Path(
-            sanitize_filepath(
-                os.path.join(Path.home(), "dagshub", "models", self.repo_api.full_name)
-            )
-        )
+        return Path(sanitize_filepath(os.path.join(Path.home(), "dagshub", "models", self.repo_api.full_name)))
 
     @cached_property
     def repo_storages(self) -> List[StorageAPIEntry]:
@@ -99,7 +95,7 @@ class ModelLocator:
     def _handle_path(self, path: str) -> Tuple[str, StorageType]:
         """
         Handles user-printed path.
-        Returns a resulting path that can be queried from DagsHub + boolean for whether it's a repo or a bucket path
+        Returns a resulting path that can be queried from DagsHub + enum for whether it's a repo or a bucket path
         If path is a repo path, returns (path as is, StorageType.Repo)
         If path is a storage path, ({actual path in storage} + StorageType.Bucket/DagshubStorage)
         Storage path resolution is based on bucket name, if the path starts with a name of an existing bucket,
@@ -113,7 +109,7 @@ class ModelLocator:
         if str_path.startswith("dagshub_storage/"):
             return str_path[len("dagshub_storage/") :], StorageType.DagshubStorage
         for storage in self.repo_storages:
-            if str_path.startswith(f"{storage.name}/"):
+            if str_path.startswith(f"{storage.name}"):
                 bucketPath = f"{storage.protocol}/{str_path}"
                 return bucketPath, StorageType.Bucket
         return path, StorageType.Repo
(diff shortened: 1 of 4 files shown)


EvgeniLeonti (commented):
LGTM; Small comment about (probably) redundant comment change

dagshub/models/model_locator.py
```diff
@@ -99,7 +95,7 @@ def repo_storages(self) -> List[StorageAPIEntry]:
     def _handle_path(self, path: str) -> Tuple[str, StorageType]:
         """
         Handles user-printed path.
-        Returns a resulting path that can be queried from DagsHub + boolean for whether it's a repo or a bucket path
+        Returns a resulting path that can be queried from DagsHub + enum for whether it's a repo or a bucket path
```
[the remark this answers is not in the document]
kbolashev: Yea, I don't really love how that docstring is worded.
Rewrote it a bit differently, lmk if you like it more.

dagshub/models/model_locator.py
```diff
@@ -113,7 +109,7 @@ def _handle_path(self, path: str) -> Tuple[str, StorageType]:
         if str_path.startswith("dagshub_storage/"):
             return str_path[len("dagshub_storage/") :], StorageType.DagshubStorage
         for storage in self.repo_storages:
-            if str_path.startswith(f"{storage.name}/"):
+            if str_path.startswith(f"{storage.name}"):
```
guysmoilov, line 112: This is actually also wrong if something like this happens:
```python
str_path = a/b/c/def
storage.name = a/b/c/d
```
I think you need to make sure they both end with a slash and compare that

tests/model_loading/test_locate.py
```diff
@@ -58,6 +58,16 @@ def test_found_when_passed_path(repo_mock):
     assert loader.path == PosixPath(dir_path)
 
 
+def test_found_when_passed_bucket_root_as_path(repo_mock):
+    bucket_name = "some-bucket/random/prefix"
+    repo_mock.add_storage("s3", bucket_name)
+    repo_mock.add_storage_file(f"s3/{bucket_name}/model.pt", b"blablabla")
+    locator = ModelLocator(repo_mock, path=bucket_name)
```
guysmoilov, line 65: To see the bug, you can try this additional test case
`path=bucket_name + "bla"`

guysmoilov (changes requested):

dagshub/models/model_locator.py
kbolashev: Ok, I get the bug now, thanks for catching it
I don't think forcing a slash is a good idea (will break cases where path points to a yaml), so I'll check for either a prefix with a slash, or a full match

dagshub/models/model_locator.py
```diff
@@ -108,8 +109,10 @@ def _handle_path(self, path: str) -> Tuple[str, StorageType]:
 
         if str_path.startswith("dagshub_storage/"):
             return str_path[len("dagshub_storage/") :], StorageType.DagshubStorage
+        in_path = PurePosixPath(str_path)
         for storage in self.repo_storages:
-            if str_path.startswith(f"{storage.name}"):
+            storage_path = PurePosixPath(storage.name)
+            if is_path_relative_to(in_path, storage_path):
```
guysmoilov, line 115: I don't fully understand how this logic relates to the simpler older one of `startswith` - did you check that it works correctly when the supplied path is exactly the root of the repo? Will `relative_to` return `.` in that case and work correctly?

guysmoilov (approved):

dagshub/models/model_locator.py
kbolashev: I don't really care about _what_ the relative path exactly is, I just care that in_path is a sub-path of the storage path.

Just in case, went to check this right now:
```
>>> a = PurePosixPath("/")
>>> b = PurePosixPath("/")
>>> a.is_relative_to(b)
True
```

So we should be fine, if a and b is the same it is considered relative

PR #566, opened 2024-12-18
Feature: Datapoint version history

kbolashev:
Add a function `Datapoint.get_version_timestamps()` that allows to fetch version history from the backend.

Signature + docs of the function:
```python
def get_version_timestamps(
    self,
    fields: Optional[List[str]] = None,
    from_time: Optional[datetime.datetime] = None,
    to_time: Optional[datetime.datetime] = None,
) -> List[DatapointHistoryResult]:
    """
    Get the timestamps of all versions of this datapoint, where the specified fields have changed.

    Args:
        fields: List of fields to check for changes. If None, all fields are checked.
        from_time: Only search versions since this time. If None, the start time is unbounded
        to_time: Only search versions until this time. If None, the end time is unbounded

    Returns:
        List of objects with information about the versions.
    """
```

I also changed the introspection code a little bit. Made it so the introspections are now cached per-host for the whole python lifetime (and not per data client as it used to be) + made it so the validators run every time you run `cl._exec`.

### DIFF
--- a/dagshub/data_engine/client/data_client.py
+++ b/dagshub/data_engine/client/data_client.py
@@ -1,3 +1,4 @@
+import datetime
 import logging
 from typing import Any, Optional, List, Dict, Union, TYPE_CHECKING
 
@@ -17,6 +18,7 @@ from dagshub.data_engine.client.models import (
     DatasourceResult,
     DatasetResult,
     MetadataFieldSchema,
+    DatapointHistoryResult,
 )
 from dagshub.data_engine.client.models import ScanOption
 from dagshub.data_engine.client.gql_mutations import GqlMutations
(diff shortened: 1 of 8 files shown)


dagshub/data_engine/client/data_client.py
```diff
@@ -361,11 +379,71 @@ def get_datasets(self, id: Optional[Union[str, int]], name: Optional[str]) -> Li
...
+
+    def get_datapoint_history(
+        self,
+        datapoints: List["Datapoint"],
+        fields: Optional[List[str]],
+        from_time: Optional[datetime.datetime],
+        to_time: Optional[datetime.datetime],
+    ) -> Dict[str, List[DatapointHistoryResult]]:
+        ds_id = datapoints[0].datasource.source.id
+        assert ds_id is not None
```
guysmoilov, line 399: Maybe also assert that the list isn't empty

guysmoilov (approved):
I have no idea how to review the gql related parts, but the feature itself seems OK to me

PR #574, opened 2025-01-01
Generalized MLFlow Model Loading

jinensetpal:
Imports and uses loader module based on model flavor, so client is no longer tied to pyfunc.

### DIFF
--- a/dagshub/data_engine/model/query_result.py
+++ b/dagshub/data_engine/model/query_result.py
@@ -7,7 +7,6 @@ from dataclasses import field, dataclass
 from os import PathLike
 from pathlib import Path
 from typing import TYPE_CHECKING, List, Dict, Any, Optional, Union, Tuple, Literal, Callable
-import json
 import os
 import os.path
 import importlib
@@ -63,6 +62,10 @@ class VisualizeError(Exception):
     """:meta private:"""
 
     pass
+
+
+def identity_func(x):
+    return x
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


dagshub/data_engine/model/query_result.py
```diff
@@ -490,7 +491,9 @@ def predict_with_mlflow_model(
 
         Args:
             repo: repository to extract the model from
-            name: name of the model in the mlflow registry
+            name: name of the model in the mlflow registry. \
+                    Supported model flavors: `torch`, `tensorflow`, `pyfunc`, `scikit-learn`. \
+                    Any model that uses the `self.predict` API is also supported. \
```
kbolashev, line 496: This shouldn't be under the name argument, and on the top level docs of the function instead.
Also sphinx's docs use double backticks for inline code instead of single backticks

dagshub/data_engine/model/query_result.py
```diff
@@ -517,16 +520,26 @@ def __next__(self):
...
         token = get_token()
         os.environ["MLFLOW_TRACKING_USERNAME"] = UserAPI.get_user_from_token(token).username
         os.environ["MLFLOW_TRACKING_PASSWORD"] = token
+        model_uri = f"models:/{name}/{version}"
+
         try:
-            model = mlflow.pyfunc.load_model(f"models:/{name}/{version}")
+            model_flavor = ".".join(
+                mlflow.models.get_model_info(model_uri).flavors["python_function"]["loader_module"].split(".")[:2]
+            )
```
kbolashev, line 534: This is kind of incromprehensible.
At the very least this is not a `model_flavor` because there's a dot in there. So this looks like a `loader_module`. 

I had Dean explain to me that the reason for this dance with cutting off everything but the first two elements is to get rid of the `.model` that shows up accidentally sometimes for some versions. Your solution will break the moment MLflow introduces a loader module that's nested more than 2 levels deep, so something like `mlflow.onnx_models.cool_models` for example. 

So you can probably do something like this and it would be more readable (wrote out the code without testing so it might not work right away):
```python
loader_module_elems = mlflow.models.get_model_info(model_uri).flavors["python_function"]["loader_module"].split(".")
if loader_module_elems[-1] == "model":
    loader_module_elems.pop()
loader_module = ".".join(loader_module_elems)
```

kbolashev (changes requested):

PR #578, opened 2025-01-08
Disentangled mechanism to obtain mlflow model

jinensetpal:
https://github.com/DagsHub/ls-configurable-model/ also loads mlflow models from dagshub repos, same as `{predict,annotate}_with_mlflow`, code is duplicated and changes need to be propagated. This disentaglement helps with code re-use.

### DIFF
--- a/dagshub/mlflow/get_model.py
+++ b/dagshub/mlflow/get_model.py
@@ -1,0 +1,48 @@
+from dagshub.common.util import lazy_load, multi_urljoin
+from dagshub.common.api import UserAPI
+from dagshub.auth import get_token
+
+from typing import TYPE_CHECKING
+import importlib
+import os
+
+if TYPE_CHECKING:
+    import mlflow
+else:
+    mlflow = lazy_load("mlflow")
+
+
+def get_mlflow_model(repo, name, host, version):
+    """
+    Get MLflow Model from the specified DagsHub repository, \
+            patched to forward the primary prediction function to `predict`.
+
+    Args:
+        repo: repository to extract the model from
+        name: name of the model in the repository's MLflow registry.
+        host: address of the DagsHub instance with the repo to load the model from.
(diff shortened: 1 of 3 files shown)


dagshub/mlflow/get_model.py
```diff
@@ -0,0 +1,48 @@
...
+if TYPE_CHECKING:
+    import mlflow
+else:
+    mlflow = lazy_load("mlflow")
+
+
+def get_mlflow_model(repo, name, host, version):
+    """
+    Get MLflow Model from the specified DagsHub repository, \
+            patched to forward the primary prediction function to `predict`.
```
kbolashev, line 18: 1) Add typehints on everything + sane defaults if you want this to be a public facing function
2) the backslash for carrying over is redundant, rst is basically the same as markdown, where you only get a newline in the generated documentation if you do a double-newline
3) The "patched to forward the primary prediction function to `predict`" sentence doesn't really convey anything (to me at least), so maybe it can just be dropped? 

Let me know what you think about this wording for the docstring:

> Load an MLflow registered model from the specified DagsHub repository.
> Call `model.predict(<data>)` on the loaded model to infer on your data.

Maybe it could be nice to have example code with a model we could point at, but I think the code for setting up the data might be too unwieldy.

kbolashev (approved):
Great idea, I think it could be a useful public-facing function!

dagshub/mlflow/get_model.py
jinensetpal: Yep makes sense! I like the new wording, the reason we mentioned it explicitly is because the MLflow API is a bit inconsistent with each model flavor, and we are overriding some behavior to make it consistent. Your bit conveys that clearer imo.

jinensetpal: Also did manage to get example code, it was YOLOv8 directly accepts file URLs so it was pretty convenient.

PR #579, opened 2025-01-08
Made LS Cient requests tenacious

jinensetpal:
+ fixed incorrect function call

If the DH LS instance shuts down due to inactivity but the client variable persists, the client does not retry queries until the instance is back online. This changes overrides the request mechanism to make requests tenacious.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -7,3 +7,4 @@ pytest-mock==3.14.0
 fiftyone==0.23.8
 datasets==2.19.1
 ultralytics==8.3.47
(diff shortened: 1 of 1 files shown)


jinensetpal:
Found a bug fixing now

kbolashev:
Another design, after talk with Guy (implement myself), that won't force us to fix it after every new version:
- Wrap all callables/properties with a call to a "is workspace alive" endpoint
- make sure only the top-level stack call is doing that
- Have a global throttle so the "is workplace alive" check is run only once every x seconds

PR #583, opened 2025-01-21
Bug: Can't upload single file to the repo bucket

kbolashev:
Repoduction:
```
$ dagshub upload user/repo a.txt --bucket
```

Ends up throwing out an exception about s3 client not being able to handle PurePosixPath. This PR fixes it by casting to string at the point of uploading to S3

Along the way figured out that single file uploading test also wasn't really testing the path correctly, so fixed that too.

### DIFF
--- a/dagshub/upload/wrapper.py
+++ b/dagshub/upload/wrapper.py
@@ -6,7 +6,6 @@ import posixpath
 import time
 import urllib
 from http import HTTPStatus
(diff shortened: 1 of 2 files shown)


kbolashev:
Also fixed:
- Can't upload single files to regular repos too (oops)
- Trying to upload the current local directory crashes

PR #589, opened 2025-02-05
Create gen-pydoc.py

guysmoilov:
For consumption by LLMs.

### DIFF
--- a/gen-pydoc.py
+++ b/gen-pydoc.py
@@ -1,0 +1,30 @@
+import pydoc
+import pkgutil
+import importlib
+import sys
+
+def generate_docs(package_name):
+    package = importlib.import_module(package_name)
+    print(f"\n# pydoc documentation for package: {package_name}\n")
+    print("\n---\n")
+    
+    for _, module_name, is_pkg in pkgutil.walk_packages(package.__path__, package.__name__ + "."):
+        print(f"\n## pydoc of module: {module_name}\n")
+        print("\n---\n")
+        
+        try:
+            docstring = pydoc.plain(pydoc.render_doc(module_name))
+            print(docstring)
+        except Exception as e:
+            print(f"Error documenting {module_name}: {e}\n")
+        
+        print("\n---\n")
+        print(f"\nEnd of documentation for module: {module_name}\n")
+
+if __name__ == "__main__":
+    if len(sys.argv) != 2:
+        print("Usage: python gen-pydoc.py <package_name>")
+        sys.exit(1)
+    
+    package_name = sys.argv[1]
+    generate_docs(package_name)
--- a/gen-pydoc.py
+++ b/gen-pydoc.py
@@ -1,30 +1,45 @@
 import pydoc
 import pkgutil
 import importlib
+import os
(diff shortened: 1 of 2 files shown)


gen-pydoc.py
```diff
@@ -0,0 +1,45 @@
...
+    out_f.write(f"\n# pydoc documentation for package: {package_name}\n")
+    out_f.write("\n---\n")
+    
+    for _, module_name, is_pkg in pkgutil.walk_packages(package.__path__, package.__name__ + "."):
+        out_f.write(f"\n## pydoc of module: {module_name}\n")
+        out_f.write("\n---\n")
+        
+        try:
+            docstring = pydoc.plain(pydoc.render_doc(module_name))
+            out_f.write(docstring)
```
kbolashev, line 22: This will ignore all of the `:meta private` tags I have on private functions that users aren't supposed to be using
Don't care about it that much, but I am concerned of LLMs abusing private things in the library.
Remains to be seen i guess

gen-pydoc.py
```diff
@@ -0,0 +1,45 @@
...
+    package = importlib.import_module(package_name)
+    
+    out_f = open(output_file, 'w') if output_file else sys.stdout
+    
+    out_f.write(f"\n# pydoc documentation for package: {package_name}\n")
+    out_f.write("\n---\n")
+    
+    for _, module_name, is_pkg in pkgutil.walk_packages(package.__path__, package.__name__ + "."):
+        out_f.write(f"\n## pydoc of module: {module_name}\n")
+        out_f.write("\n---\n")
```
kbolashev, line 18: Did you check how it works in cases where there's a module-level import from another module?
e.g. most of the things listed here: https://github.com/DagsHub/client/blob/master/dagshub/__init__.py
are imported from another module, so it could be documenting those functions two times

kbolashev (changes requested):

gen-pydoc.py
guysmoilov: Everything is printed only once when I run it on the dagshub module

gen-pydoc.py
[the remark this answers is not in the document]
guysmoilov: OK, I wasn't sure if I'm allowed to add things there or if it will mess things up

gen-pydoc.py
kbolashev: > Probably less important for LLMs

In my opinion this is actually MUCH MORE important for LLMs, because if the docs show up there, LLM will just pick it up and regurgitate to the user, and then the user ends up writing code that mangles private things that shouldn't be mangled.

kbolashev (approved):
Fix the lint (you can run `ruff fmt` or `black`) and feel free to merge

gen-pydoc.py
guysmoilov: Will try to find a possible solution, probably not easy.

kbolashev: I don't think there's a point in worrying about it yet, and if we have users complaining about these functions/fields changing we can tell them that they're intended to be private.

PR #590, opened 2025-02-06
Feature: Limits as part of the query

kbolashev:
This PR adds ability to make limits a part of a query with `ds.limit(n)`. 
This allows to save the limit as part of a dataset and/or an MLflow artifact.

Should allow for better reproducibility for cases where the user is using `ds.head()` to train on a subset.

This PR also adds a new `ds.fetch()` function, that respects this limiter AND also always log an MLflow artifact to an active run.

```python
q = ds.limit(100)
q.fetch()      # Gets 100 datapoints, logs to mlflow
q.head(50)     # Gets 50 datapoints, doesn't log to mlflow
q.head(500)    # Gets 100 datapoints, doesn't log to mlflow
q.all()        # Gets ALL datapoints, the user will get a warning that there was a limit and it gotten overriden
```


simonlsk (approved):
LGTM, 
Would be good to have introspection here to check if the server supports limit.

PR #595, opened 2025-02-18
Bug: refetch fields on has_field failure

kbolashev:
Users might add new fields outside of the client, that won't appear in the client immediately and could fail when `ds.has_field()` is called (in client or user code).

This PR is adding a re-fetch to `ds.has_field()` that refreshes the datasource state if the field is not found.


simonlsk (approved):
LGTM, 
Thanks

PR #600, opened 2025-02-27
Fix local auto labeling

deanp70:
Fixed an issue where the auto-labeling expected format for local and dockerized instances was different.

Need to update the docs [active learning](https://dagshub.com/docs/use_cases/active_learning/) page

### DIFF
--- a/dagshub/data_engine/model/query_result.py
+++ b/dagshub/data_engine/model/query_result.py
@@ -963,7 +963,7 @@ class QueryResult:
                 _Batcher(dset, batch_size) if batch_size != 1 else dset
             ):  # encapsulates dataset with batcher if necessary and iterates over it
                 for prediction, remote_path in zip(
-                    predict_fn(local_paths),
+                    [predict_fn(local_paths)],
                     [result.path for result in self[idx * batch_size : (idx + 1) * batch_size]],
                 ):
                     predictions[remote_path] = prediction
--- a/dagshub/data_engine/model/query_result.py
+++ b/dagshub/data_engine/model/query_result.py
@@ -866,7 +866,7 @@ class QueryResult:
     @staticmethod
     def _get_predict_dict(predictions, remote_path, log_to_field, is_prediction=False):
         ls_json_key = "annotations" if not is_prediction else "predictions"
-        res = {log_to_field: json.dumps({ls_json_key: predictions[remote_path][0]}).encode("utf-8")}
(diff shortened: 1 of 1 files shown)


dagshub/data_engine/model/query_result.py
```diff
@@ -548,6 +549,7 @@ def predict_with_mlflow_model(
         post_hook: Callable[[Any], Any] = identity_func,
         batch_size: int = 1,
         log_to_field: Optional[str] = None,
+        is_prediction: Optional[bool] = True
     ) -> Dict[str, Any]:
```
kbolashev, line 553: Having a `is_prediction` kwarg on a `predict_with_...` function is IMO really confusing API design.
I won't put my foot down and I will push this through for whatever you need right now, but we will probably need to change this in the near future if you want people to actually be using this.

Either way, thanks for adding the docstring

dagshub/data_engine/model/query_result.py
```diff
@@ -956,7 +963,7 @@ def generate_predictions(
                 _Batcher(dset, batch_size) if batch_size != 1 else dset
             ):  # encapsulates dataset with batcher if necessary and iterates over it
                 for prediction, remote_path in zip(
-                    predict_fn(local_paths),
+                    [predict_fn(local_paths)],
                     [result.path for result in self[idx * batch_size : (idx + 1) * batch_size]],
```
kbolashev, line 967: [blocker] This doesn't look right.
The signature of `CustomPredictor` is:
```python
CustomPredictor = Callable[
    [List[str]],
    List[Tuple[Any, Optional[float]]],
]
```

So the function should be accepting a list of datapoints, and return a list of prediction tuples. 
Why are you wrapping the result in another list? This zip will end up processing all of the returned predictions for one single path.

kbolashev: How does the return from the prediction function you're using right now look like?

dagshub/data_engine/model/query_result.py
kbolashev: Extracted the main code into a `ds._infer_with_mlflow_on_files`, gotten rid of the kwarg in the original functions and instead just call this extracted function

dagshub/data_engine/model/query_result.py
kbolashev: Apparently this is happening only when batch size = 1. Bigger batch sizes are supposed to work correctly.
Honestly don't know why not just use the Batcher even for batch size = 1, but I don't want to introduce too many changing variables right now, considering I can't test it that well.

PR #636, opened 2025-12-14
Bug: Don't parse annotations as documents, even if the tag is set

kbolashev:
In Data Engine, if you have a metadata field that is set both as an annotation and a document, loading the annotation field fails.

Example of setting an annotation AND a document (might happen automatically on the backend):
```python
    builder = ds.metadata_field("exported_annotation").set_annotation()
    builder._add_tags({ReservedTags.DOCUMENT.value})
    builder.apply()
```

This PR fixes it, now if an annotation field also has the document field set, the annotation tag takes precedence, and it's ignored for loading documents.


dagshub/data_engine/model/query_result.py
```diff
@@ -418,6 +419,8 @@ def _get_blob_fn(dp: Datapoint, field: str, url: str, blob_path: Path):
 
         # Convert any downloaded document fields
         document_fields = [f for f in fields if f in self.document_fields]
+        # Exclude any annotation fields, because they are already converted above
+        document_fields = [f for f in document_fields if f not in self.annotation_fields]
```
simonlsk, line 423: It would make more sense to me for every processed field to either be loaded as document or annotation here in this function, and not called twice from outside and then filter out annotations when it's called for documents.

In the current flow if we were to add a new tag that conflicts with the existing one, you would need to double exclude every field from the ones they don't belong to.

Let me know if this fix is easy to make.

simonlsk (approved):
Seems to solve the issue thank you.
I have a suggestion to make the flow more solid, but otherwise LGTM.

dagshub/data_engine/model/query_result.py
kbolashev: Would require a refactor, due to the way annotation errors are handled and printed out
Can totally be refactored though, just not a 5 minute refactor

PR #645, opened 2026-01-25
Add support for W3C Trace Context in graphql requests

guysmoilov:
Also catch and wrap TransportServerError to add the trace ID in the logged error.

### DIFF
--- a/dagshub/data_engine/client/data_client.py
+++ b/dagshub/data_engine/client/data_client.py
@@ -185,7 +185,7 @@ class DataClient:
         if validate:
             query.validate_params(params if params else {}, self.query_introspection)
         q = gql.gql(query.generate())
-        headers = config.requests_headers
+        headers = dict(config.requests_headers)
         traceparent = None
         if not config.disable_traceparent:
             traceparent = build_traceparent()


dagshub/common/tracing.py
```diff
@@ -0,0 +1,26 @@
...
+def build_traceparent() -> str:
+    """
+    Build a W3C Trace Context traceparent header value.
+
+    Format: version(2)-trace-id(32)-parent-id(16)-flags(2)
+    """
+    version = "00"
+    trace_id = _generate_non_zero_hex(16)
+    parent_id = _generate_non_zero_hex(8)
+    flags = "01"  # sampled
```
kbolashev, line 25: sampled from what?

dagshub/common/tracing.py
```diff
@@ -0,0 +1,26 @@
...
+def _generate_non_zero_hex(byte_count: int) -> str:
+    """
+    Generate a non-zero hex string of the given byte length.
+
+    Per W3C Trace Context, trace-id and parent-id MUST NOT be all zeros.
+    """
+    while True:
+        value = secrets.token_hex(byte_count)
+        if int(value, 16) != 0:
+            return value
```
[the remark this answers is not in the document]
guysmoilov: Overkill, YAGNI

dagshub/data_engine/client/data_client.py
```diff
@@ -184,10 +185,18 @@ def _exec(
...
+        headers = dict(config.requests_headers)
+        traceparent = None
+        if not config.disable_traceparent:
+            traceparent = build_traceparent()
+            headers["traceparent"] = traceparent
         try:
-            resp = self.client.execute(q, variable_values=params)
-        except TransportQueryError as e:
-            raise DataEngineGqlError(e, self.client.transport.response_headers.get("X-DagsHub-Support-Id"))
+            resp = self.client.execute(q, variable_values=params, extra_args={'headers': headers})
```
kbolashev, line 194: [offtopic kind of] The extra_args is extremely arcane and I couldn't actually find where that got passed into, had to fire up wireshark to see that the headers get actually added

dagshub/common/tracing.py
guysmoilov: RFC flag, as we discussed - setting it to "00" causes the server to not record anything by default.

dagshub/data_engine/client/data_client.py
guysmoilov: Yes, it's multiple kwargs deep, but there isn't a better way to pass headers without modifying global state or creating a new client or transport object per request, which loses session pooling.

PR #678, opened 2026-03-25
Add MOT and CVAT video annotation import/export support

deanp70:
Based on changes to dagshub-annotation-converter to the same effect: https://github.com/DagsHub/dagshub-annotation-converter/pull/25

## Implementation Details

The PR extends annotation support through a multi-layered approach:

**Importer Module Enhancements** (`importer.py`):
- Extended `AnnotationType` literal to include "mot", and "cvat_video" 
- Introduced `is_video_format` property to identify video-based annotation types
- Added video-specific annotation handling via `_flatten_video_annotations` to convert per-frame annotations into single video-entry dictionaries
- Implemented video conversion routing in `convert_to_ls_tasks` for Label Studio video task generation
- Adjusted `remap_annotations` to handle cases where video format annotations lack explicit filenames
- New internal helpers: `_is_video_annotation_dict`, `_flatten_cvat_fs_annotations`, `_flatten_mot_fs_annotations`

**Metadata Annotations Support** (`metadata.py`):
- Added `VideoRectangleAnnotation` support in Label Studio task lookup registry

**Query Result Export API** (`query_result.py`):
- Two new export methods:
  - `export_as_mot()` - supports single/multi-video export with optional dimension parameters
  - `export_as_cvat_video()` - handles video CVAT XML generation with video name customization
- Added video annotation helpers: `_get_all_video_annotations`, `_prepare_video_file_for_export`, `_get_annotation_filename`
- Introduced `_resolve_annotation_field` to auto-select available annotation field with alphabetical fallback

**Test Coverage**:
- `test_mot.py` (383 lines): Video annotation flattening, directory/zip import, MOT export structure, dimension handling
- `test_cvat_video.py` (276 lines): CVAT video XML import/export, track/box element generation, multi-datapoint support

**Code Quality and Testing Infrastructure**:
- Enhanced `ruff.toml` linting configuration with comprehensive rule set including:
  - Security checks (flake8-bandit)
  - Bug detection (flake8-bugbear)
  - Try/except pattern validation (tryceratops)
  - Unused argument detection and other design pattern checks
  - Per-file exceptions for `__init__.py` and `conftest.py`
- Added `DatasourceType.REPOSITORY` assignment in test fixtures (`tests/data_engine/conftest.py`)
- Added `id` property to `MockRepoAPI` test utility
- Minor refactor in `token_auth.py` (no behavioral change - logical equivalence in boolean expression)

**Documentation Note**: Per PR comments, the supported formats list at dagshub.com/docs/use_cases/annotation/#supported-formats-for-annotation_imports requires manual update to reflect MOT, CVAT Video support.


dagshub/__init__.py
kbolashev, line 1: Do not bump the version in PRs

dagshub/data_engine/annotation/metadata.py
kbolashev, line 117: Need to handle video annotations here too now - check what types of annotations are there and set corresponding value in the data dict.

Go through every annotation, check what it's a subclass of, if it's an image - set data[image], if it's video - set data[video]. It's ok to set both of them too!

deanp70: done

dagshub/data_engine/annotation/importer.py
```diff
@@ -78,21 +91,190 @@ def import_annotations(self) -> Mapping[str, Sequence[IRAnnotationBase]]:
...
+        if video_name is None:
+            video_name = self.annotations_file.stem
+
+        all_anns: List[IRTaskAnnotation] = []
+        for frame_anns in video_data.values():
+            all_anns.extend(frame_anns)
+        return {video_name: all_anns}
+
+    def _flatten_cvat_fs_annotations(
+        self, fs_annotations: Mapping[str, object]
```
kbolashev: Need the value type of `fs_annotations` to be `CVATAnnotations`

dagshub/data_engine/annotation/importer.py
```diff
@@ -78,21 +91,190 @@ def import_annotations(self) -> Mapping[str, Sequence[IRAnnotationBase]]:
...
+                        flattened[video_key].extend(frame_anns)
+                else:
+                    for filename, anns in result.items():
+                        flattened.setdefault(filename, [])
+                        flattened[filename].extend(anns)
+        return flattened
+
+    def _flatten_mot_fs_annotations(
+        self,
+        fs_annotations: Mapping[str, object],
```
kbolashev: Need the value type of `fs_annotations` to be `Tuple[IRVideoSequence, MOTContext]`

dagshub/data_engine/annotation/importer.py
```diff
@@ -78,21 +91,190 @@ def import_annotations(self) -> Mapping[str, Sequence[IRAnnotationBase]]:
...
+        if isinstance(result, IRVideoSequence):
+            return True
+        if not isinstance(result, dict) or len(result) == 0:
+            return False
+        first_key = next(iter(result.keys()))
+        return isinstance(first_key, int)
+
+    def _flatten_video_annotations(
+        self,
+        video_data,
```
kbolashev: Add type here
Seems like it's `Union[IRVideoSequence, CVATAnnotations]` based on what I'm inferring from usages

kbolashev: Also I don't see what's "flattening" about this function and all the other `_flatten_...` functions, seems like their point is to get the path of the datapoint associated with the annotation

dagshub/data_engine/annotation/importer.py
```diff
@@ -78,21 +91,190 @@ def import_annotations(self) -> Mapping[str, Sequence[IRAnnotationBase]]:
...
+                if self._is_video_annotation(result):
+                    video_key = self._first_video_annotation_filename(result.values())
+                    video_key = self._resolve_video_annotation_key(video_key, fallback=str(rel_path))
+                    flattened.setdefault(video_key, [])
+                    for frame_anns in result.values():
+                        flattened[video_key].extend(frame_anns)
+                else:
+                    for filename, anns in result.items():
+                        flattened.setdefault(filename, [])
+                        flattened[filename].extend(anns)
```
kbolashev: In this and all the other flatten functions - what is this dict typecheck? Can it actually happen?
dict is not IRTaskAnnotation, so how can you go over `fs_annotations` and get dicts there?
Either the input type needs to be expanded to Union[IRTaskAnnotation, dict], or this else needs to be removed

dagshub/data_engine/annotation/importer.py
```diff
@@ -78,21 +91,190 @@ def import_annotations(self) -> Mapping[str, Sequence[IRAnnotationBase]]:
...
+            else:
+                raise ValueError(f"Unsupported annotation type: {self.annotations_type}")
 
             return annotation_dict
 
+    @staticmethod
+    def _is_video_annotation(result) -> bool:
+        """Check whether a loader result contains video annotations."""
+        from dagshub_annotation_converter.ir.video import IRVideoSequence
+
```
kbolashev: no inline imports

dagshub/data_engine/annotation/importer.py
kbolashev: Digging deeper, I see that it's because `CVATImageAnnotations = Dict[str, Sequence[IRImageAnnotationBase]]`, and it's part of the union of `CVATAnnotation`, so that's why it's checking for a dict.
Write that in the `elif` at least.
Also, if it IS actually a dict, then what we're handling is an image annotation and not video annotation, therefore is there a point in even handling these?

dagshub/data_engine/annotation/importer.py
```diff
@@ -78,21 +91,190 @@ def import_annotations(self) -> Mapping[str, Sequence[IRAnnotationBase]]:
...
+        return self.annotations_file.stem
+
+    @classmethod
+    def _first_video_annotation_filename(
+        cls,
+        frame_groups: Iterable[Sequence[IRAnnotationBase]],
+    ) -> Optional[str]:
+        for frame_anns in frame_groups:
+            for ann in frame_anns:
+                ann_filename = cls._stringify_video_path(getattr(ann, "filename", None))
```
kbolashev: If the input is really `Iterable[Sequence[IRAnnotationBase]]`, then the filename attribute will always be there

dagshub/data_engine/annotation/importer.py
```diff
@@ -78,21 +91,190 @@ def import_annotations(self) -> Mapping[str, Sequence[IRAnnotationBase]]:
...
+                    return ann_filename
+        return None
+
     def download_annotations(self, dest_dir: Path):
         log_message("Downloading annotations from repository")
         repoApi = self.ds.source.repoApi
-        if self.annotations_type == "cvat":
-            # Download just the annotation file
+        if self.annotations_type in ("cvat", "cvat_video"):
             repoApi.download(self.annotations_file.as_posix(), dest_dir, keep_source_prefix=True)
```
kbolashev, line 263: Bring back the comment

dagshub/data_engine/annotation/importer.py
```diff
@@ -104,6 +286,8 @@ def download_annotations(self, dest_dir: Path):
             # Download the annotation data
             assert context.path is not None
             repoApi.download(self.annotations_file.parent / context.path, dest_dir, keep_source_prefix=True)
+        elif self.annotations_type == "mot":
+            repoApi.download(self.annotations_file.as_posix(), dest_dir, keep_source_prefix=True)
```
kbolashev, line 275: Calls the same function as the cvat calls, could be put in there for DRY, although I can see how it might be nicer-looking for "pattern matching" to keep it like this

dagshub/data_engine/annotation/importer.py
```diff
@@ -78,21 +91,190 @@ def import_annotations(self) -> Mapping[str, Sequence[IRAnnotationBase]]:
...
+                    else:
+                        annotation_dict = result
+            elif self.annotations_type == "mot":
+                mot_kwargs = {}
+                if "image_width" in self.additional_args:
+                    mot_kwargs["image_width"] = self.additional_args["image_width"]
+                if "image_height" in self.additional_args:
+                    mot_kwargs["image_height"] = self.additional_args["image_height"]
+                if "video_name" in self.additional_args:
+                    mot_kwargs["video_file"] = self.additional_args["video_name"]
```
kbolashev, line 120: Everything in the `additional_args` dict needs to be documented in the docstring of `dagshub.data_engine.model.datasource.Datasource.import_annotations_from_files`

dagshub/data_engine/annotation/importer.py
```diff
@@ -78,21 +91,190 @@ def import_annotations(self) -> Mapping[str, Sequence[IRAnnotationBase]]:
...
+    def _is_video_annotation(result) -> bool:
+        """Check whether a loader result contains video annotations."""
+        from dagshub_annotation_converter.ir.video import IRVideoSequence
+
+        if isinstance(result, IRVideoSequence):
+            return True
+        if not isinstance(result, dict) or len(result) == 0:
+            return False
+        first_key = next(iter(result.keys()))
+        return isinstance(first_key, int)
```
kbolashev: This function is called for cvat only, so the interface/signature for it should be something like 
`def _determine_cvat_annotation(result: CVATAnnotation) -> Literal["video", "image"]`

dagshub/data_engine/annotation/importer.py
```diff
@@ -153,8 +340,20 @@ def remap_annotations(
                 )
                 continue
             for ann in anns:
-                assert ann.filename is not None
-                ann.filename = remap_func(ann.filename)
+                if isinstance(ann, IRVideoAnnotationTrack):
+                    for track_ann in ann.annotations:
+                        if track_ann.filename is not None:
+                            track_ann.filename = remap_func(track_ann.filename)
```
kbolashev: Considering that all annotations in a track are supposed to be connected to one file, this can probably be skipped and it should always be set to `new_filename`.
Might even be better for consistency

dagshub/data_engine/annotation/importer.py
```diff
@@ -296,3 +497,38 @@ def convert_to_ls_tasks(self, annotations: Mapping[str, Sequence[IRAnnotationBas
...
+            if ls_task is not None:
+                tasks[filename] = ls_task.model_dump_json().encode("utf-8")
+        return tasks
+
+    @staticmethod
+    def _build_video_sequence(
+        annotations: Sequence[IRTaskAnnotation],
+        filename: str,
+    ) -> Optional[IRVideoSequence]:
+        tracks = [ann.model_copy(deep=True) for ann in annotations if isinstance(ann, IRVideoAnnotationTrack)]
```
kbolashev: Why copy them? Seems like unnecessary memory churn

dagshub/data_engine/annotation/metadata.py
```diff
@@ -22,6 +24,11 @@
 
     from dagshub.data_engine.model.datapoint import Datapoint
 
+from dagshub_annotation_converter.formats.label_studio.videorectangle import VideoRectangleAnnotation
+from dagshub_annotation_converter.formats.label_studio.task import task_lookup as _task_lookup
+
+_task_lookup["videorectangle"] = VideoRectangleAnnotation
+
```
kbolashev: Remove this

dagshub/data_engine/annotation/video.py
```diff
@@ -0,0 +1,45 @@
...
+    # Pre-group annotations into tracks (required by new from_annotations API)
+    by_track: dict[str, list[IRVideoBBoxFrameAnnotation]] = defaultdict(list)
+    for ann in annotations:
+        object_id = getattr(ann, "object_id", None)
+        if object_id is None and ann.__pydantic_extra__ is not None:
+            object_id = ann.__pydantic_extra__.get("object_id")
+        if object_id is None:
+            object_id = ann.imported_id
+        if object_id is None:
+            raise ValueError("Video annotation is missing an object identifier")
```
kbolashev: Crazy chunk of checks, i don't understand a thing here.
I understand the idea that you want to get the object_id and trying to get it from wherever you can in the annotation, but these all seem weird.

dagshub/data_engine/model/query_result.py
```diff
@@ -778,6 +784,88 @@ def _get_all_annotations(self, annotation_field: str) -> List[IRImageAnnotationB
...
+    def _prepare_video_file_for_export(self, local_root: Path, repo_relative_filename: str) -> Optional[Path]:
+        ann_path = Path(repo_relative_filename)
+        primary = local_root / ann_path
+        if primary.exists():
+            return primary
+        source_prefix = Path(self.datasource.source.source_prefix)
+        with_prefix = local_root / source_prefix / ann_path
+        if with_prefix.exists():
+            return with_prefix
+        return None
```
kbolashev: There's nothing about this that's preparing anything, it just checks if a path exist. Nothing about it is related to video either.
Please come up with a better name for a function and put a docstring on what it's supposed to be doing

dagshub/data_engine/model/query_result.py
```diff
@@ -778,6 +784,88 @@ def _get_all_annotations(self, annotation_field: str) -> List[IRImageAnnotationB
...
+
+    def _get_media_export_layout(self, download_dir: Path, media_dir_name: str) -> Tuple[Path, Path, Path]:
```
kbolashev: Same comment w.r.t. docstring

kbolashev: My brain is being wrecked by these last couple of functions :(

dagshub/data_engine/model/query_result.py
```diff
@@ -778,6 +784,88 @@ def _get_all_annotations(self, annotation_field: str) -> List[IRImageAnnotationB
...
+    def _get_annotation_filename(ann: IRVideoBBoxFrameAnnotation) -> Optional[str]:
+        filename = ann.filename
+        if filename is None:
+            return None
+        if isinstance(filename, (list, tuple)):
+            if len(filename) == 0:
+                return None
+            if len(filename) > 1:
+                raise ValueError(f"Annotation has multiple filenames: {filename}")
+            filename = filename[0]
```
kbolashev: Filename is `Optional[str]`, it can't be either of list or tuple. How did this show up here?

dagshub/data_engine/model/query_result.py
```diff
@@ -778,6 +784,88 @@ def _get_all_annotations(self, annotation_field: str) -> List[IRImageAnnotationB
...
+            media_dir = media_dir.joinpath(*prefix_parts)
+        if not prefix_parts or prefix_parts[-1] != media_dir_name:
+            media_dir = media_dir / media_dir_name
+
+        dataset_root = media_dir.parent
+        labels_dir = dataset_root / "labels"
+        return media_dir, labels_dir, dataset_root
+
+    @staticmethod
+    def _get_annotation_filename(ann: IRVideoBBoxFrameAnnotation) -> Optional[str]:
```
kbolashev: Why so strict with the type? I think most of these functions can accept `IRVideoFrameAnnotationBase` at the very least.

dagshub/data_engine/model/query_result.py
```diff
@@ -861,6 +944,195 @@ def export_as_yolo(
...
+            {
+                Path(ann_filename).name
+                for ann_filename in (self._get_annotation_filename(ann) for ann in video_annotations)
+                if ann_filename
+            }
+        )
+        has_multiple_sources = len(source_names) > 1
+
+        log_message(f"Downloading videos into {video_dir}...")
+        local_download_root = self.download_files(video_dir, keep_source_prefix=False)
```
kbolashev: you should probably only download files that have annotations. Right now all of the files in the query result will be downloaded

dagshub/data_engine/model/query_result.py
```diff
@@ -861,6 +944,195 @@ def export_as_yolo(
...
+            context.video_width = image_width
+            context.video_height = image_height
+            export_mot_sequences_to_dirs(sequences, context, dataset_root)
+            result_path = dataset_root
+        else:
+            video_file: Optional[Path] = None
+            ref_filename = next((self._get_annotation_filename(a) for a in video_annotations), None)
+            if ref_filename is None:
+                raise FileNotFoundError("Missing annotation filename for MOT export.")
+            video_file = self._prepare_video_file_for_export(local_download_root, ref_filename)
```
kbolashev: Don't understand this. You might as well just do a `video_annotations[0].filename` here

kbolashev (changes requested):

dagshub/data_engine/annotation/video.py
deanp70: Fixed and simplified

dagshub/data_engine/annotation/importer.py
deanp70: OK, done.

dagshub/data_engine/annotation/importer.py
deanp70: done

dagshub/data_engine/annotation/importer.py
deanp70: Since CVATAnnotations includes IRVideoSequence we just have CVATAnnotations IMO. With respect to naming, I can change to _resolve_annotations if you think it's better?

dagshub/data_engine/annotation/importer.py
deanp70: Some of this (the is_video_annotation branch) is dead code, removed that. Also a bunch of dead code in _flatten_mot_fs_annotations is removed. The function is called for both cvat and cvat_video if I understand your question.

dagshub/data_engine/annotation/importer.py
deanp70: ok

dagshub/data_engine/annotation/importer.py
deanp70: I think they are logically different and doing the same download call as cvat and cvat_video is coincidental + the pattern matching argument. Let's leave as is.

dagshub/data_engine/model/query_result.py
deanp70: This is dead code, removed.

dagshub/data_engine/model/query_result.py
deanp70: Updating the whole call chain.

dagshub/data_engine/annotation/importer.py
deanp70: Will change to _key_annotation_by_filename

kbolashev:
Tests are failing

dagshub/data_engine/model/query_result.py
```diff
@@ -844,17 +856,8 @@ def _get_media_export_layout(self, download_dir: Path, media_dir_name: str) -> T
...
-            return None
-        if isinstance(filename, (list, tuple)):
-            if len(filename) == 0:
-                return None
-            if len(filename) > 1:
-                raise ValueError(f"Annotation has multiple filenames: {filename}")
-            filename = filename[0]
-        return str(filename)
+    def _get_annotation_filename(ann: IRVideoFrameAnnotationBase) -> Optional[str]:
+        return ann.filename
```
kbolashev: Can get rid of this function probably

kbolashev (commented):
one last thing I think

dagshub/data_engine/model/query_result.py
deanp70: Correct. Done.

dagshub/data_engine/model/query_result.py
```diff
@@ -988,7 +984,7 @@ def export_as_mot(
         source_names = sorted(
             {
                 Path(ann_filename).name
-                for ann_filename in (self._get_annotation_filename(ann) for ann in video_annotations)
+                for ann_filename in (ann.filename for ann in video_annotations)
                 if ann_filename
```
kbolashev: Can be simplified to use `ann.filename` everywhere

dagshub/data_engine/model/query_result.py
```diff
@@ -1067,7 +1063,7 @@ def export_as_cvat_video(
...
```
kbolashev: same thing

kbolashev (approved):

PR #679, opened 2026-03-25
Add COCO annotation import/export support

deanp70:
Based on changes to dagshub-annotation-converter to the same effect: https://github.com/DagsHub/dagshub-annotation-converter/pull/26

## Implementation Details

The PR extends annotation support through a multi-layered approach:

**Importer Module Enhancements** (`importer.py`):
- Extended `AnnotationType` literal to include "coco", 

**Metadata Annotations Support** (`metadata.py`):
- Introduced `add_coco_annotation` method to load COCO-format JSON and integrate annotations into metadata

**Query Result Export API** (`query_result.py`):
- One new export methods:
  - `export_as_coco()` - generates COCO JSON with optional class mapping and custom output filename
- Introduced `_resolve_annotation_field` to auto-select available annotation field with alphabetical fallback
- Refactored YOLO export to use the new annotation field resolution

**Test Coverage**:
- `test_coco.py` (218 lines): Import/export flows, filename rewriting, field resolution, class mapping, output customization

**Code Quality and Testing Infrastructure**:
- Enhanced `ruff.toml` linting configuration with comprehensive rule set including:
  - Security checks (flake8-bandit)
  - Bug detection (flake8-bugbear)
  - Try/except pattern validation (tryceratops)
  - Unused argument detection and other design pattern checks
  - Per-file exceptions for `__init__.py` and `conftest.py`
- Added `DatasourceType.REPOSITORY` assignment in test fixtures (`tests/data_engine/conftest.py`)
- Added `id` property to `MockRepoAPI` test utility
- Minor refactor in `token_auth.py` (no behavioral change - logical equivalence in boolean expression)

**Documentation Note**: Per PR comments, the supported formats list at dagshub.com/docs/use_cases/annotation/#supported-formats-for-annotation_imports requires manual update to reflect COCO BBOX, and COCO Segmentation support.


dagshub/auth/token_auth.py
```diff
@@ -37,7 +37,7 @@ def auth_flow(self, request: Request) -> Generator[Request, Response, None]:
 
     def can_renegotiate(self):
         # Env var tokens cannot renegotiate, every other token type can
-        return not type(self._token) is EnvVarDagshubToken
+        return type(self._token) is not EnvVarDagshubToken
```
kbolashev, line 41: what is this drive-by change?
doesn't change anything functionally, but weird

dagshub/data_engine/annotation/importer.py
```diff
@@ -153,8 +157,10 @@ def remap_annotations(
                 )
                 continue
             for ann in anns:
-                assert ann.filename is not None
-                ann.filename = remap_func(ann.filename)
+                if ann.filename is not None:
+                    ann.filename = remap_func(ann.filename)
+                else:
+                    ann.filename = new_filename
```
kbolashev: Why did you have to make this change?
I guess if the annotation already has a filename set, then you want to forcefully set it to the filename of the datapoint?

Seems a bit risky

dagshub/data_engine/annotation/importer.py
```diff
@@ -85,14 +86,15 @@ def import_annotations(self) -> Mapping[str, Sequence[IRAnnotationBase]]:
...
+            elif self.annotations_type == "coco":
+                annotation_dict, _ = load_coco_from_file(annotations_file)
 
             return annotation_dict
 
     def download_annotations(self, dest_dir: Path):
         log_message("Downloading annotations from repository")
         repoApi = self.ds.source.repoApi
         if self.annotations_type == "cvat":
-            # Download just the annotation file
```
kbolashev, line 95: Why remove the comment?
The intention of the comments was to specify what gets actually downloaded for each of the annotation types.

dagshub/data_engine/annotation/metadata.py
```diff
@@ -271,6 +271,28 @@ def add_image_pose(
...
+        self,
+        coco_json: str,
+    ):
+        """
+        Add annotations from a COCO-format JSON string.
+
+        Args:
+            coco_json: A COCO-format JSON string with ``categories``, ``images``, and ``annotations`` keys.
+        """
+        from dagshub_annotation_converter.converters.coco import load_coco_from_json_string
```
kbolashev: No inside-functions imports pls

dagshub/data_engine/annotation/metadata.py
```diff
@@ -271,6 +271,28 @@ def add_image_pose(
         self.annotations.append(ann)
         self._update_datapoint()
 
+    def add_coco_annotation(
+        self,
+        coco_json: str,
+    ):
+        """
```
kbolashev: Does this function actually make sense?
It seems like an extremely rare case that a user would have a perfectly formulated COCO JSON, including categories and images, with annotations for one single file.

dagshub/data_engine/annotation/metadata.py
```diff
@@ -271,6 +271,28 @@ def add_image_pose(
...
+            coco_json: A COCO-format JSON string with ``categories``, ``images``, and ``annotations`` keys.
+        """
+        from dagshub_annotation_converter.converters.coco import load_coco_from_json_string
+
+        grouped, _ = load_coco_from_json_string(coco_json)
+        new_anns: list[IRAnnotationBase] = []
+        for anns in grouped.values():
+            for ann in anns:
+                ann.filename = self.datapoint.path
+                new_anns.append(ann)
```
kbolashev: So you're actually overriding ALL the annotations' filenames? Seems like a footgun. What if I go and upload a 1000-count annotation JSON into one datapoint?

dagshub/data_engine/model/query_result.py
```diff
@@ -861,6 +868,54 @@ def export_as_yolo(
...
+        download_dir = Path(download_dir)
+
+        annotations = self._get_all_annotations(annotation_field)
+        if not annotations:
+            raise RuntimeError("No annotations found to export")
+
+        context = CocoContext()
+        if classes is not None:
+            context.categories = dict(classes)
+
```
kbolashev, line 907: `CocoContext.categories` is not a dictionary, you need to create a `Categories()` object

dagshub/__init__.py
```diff
@@ -1,4 +1,4 @@
-__version__ = "0.6.9"
+__version__ = "0.6.10"
```
kbolashev: [blocker] don't change versions in PRs

tests/data_engine/annotation_import/test_coco.py
```diff
@@ -0,0 +1,218 @@
...
+from dagshub.data_engine.client.models import MetadataSelectFieldSchema
+from dagshub.data_engine.dtypes import MetadataFieldType, ReservedTags
+from dagshub.data_engine.model.datapoint import Datapoint
+from dagshub.data_engine.model.query_result import QueryResult
+
+
+@pytest.fixture(autouse=True)
+def mock_source_prefix(ds):
+    with patch.object(type(ds.source), "source_prefix", new_callable=PropertyMock, return_value=PurePosixPath()):
+        yield
```
kbolashev: This should probably be in `_create_mock_datasource()` so it will work for all users of the datasource fixtures

kbolashev (changes requested):

dagshub/auth/token_auth.py
deanp70: Reverted

dagshub/data_engine/annotation/importer.py
deanp70: reverted

dagshub/data_engine/annotation/metadata.py
deanp70: fixed

dagshub/data_engine/annotation/importer.py
deanp70: reverted

dagshub/data_engine/annotation/metadata.py
deanp70: This API doesn't make sense as a datapoint-scoped method: it accepts dataset-level COCO input and collapses all parsed annotations onto one datapoint by rewriting filenames. I removed the method

PR #688, opened 2026-04-21
feat: return uploaded file URL from Repo.upload and upload_files

### LINKED ISSUE #238: Repo.upload method will return url of that uploaded file in dagshub/upload/wrapper.py
Currently repo.upload method in dagshub/upload/wrapper.py, just uploads the file and logs the error codes and its content. It would be helpful, if we can get the url of file uploaded.

singhvishalkr:
## Summary

Both `Repo.upload()` and the module-level `upload_files()` now return the DagsHub UI URL where the uploaded file(s) can be viewed.

Fixes #238

## Changes

- `Repo.upload()` returns `str` — the browsable DagsHub URL constructed via the existing `get_files_ui_url()` helper
- `upload_files()` (module-level convenience wrapper) passes through that return value
- Added `Returns` section to both docstrings

## Motivation

Currently `Repo.upload()` returns `None`, forcing callers to manually reconstruct the URL if they want to link to the uploaded content. The `get_files_ui_url()` helper was already used internally for directory upload logging — this change exposes the same URL to callers.

## Testing

Verified that `get_files_ui_url` produces the correct `{owner}/{reponame}/src/{branch}/{path}` URL by tracing the existing directory upload log message. No new dependencies introduced.

### DIFF
--- a/dagshub/upload/wrapper.py
+++ b/dagshub/upload/wrapper.py
@@ -329,7 +329,7 @@ class Repo:
             except ValueError:
                 # local_path is outside cwd, use only its basename
                 remote_path = local_path.name
-        remote_path = cast(str, Path(remote_path).as_posix())
+        remote_path = posixpath.normpath(cast(str, Path(remote_path).as_posix()))
 
         if bucket:
             self.upload_files_to_bucket(local_path, remote_path, **kwargs)
@@ -602,7 +602,7 @@ class Repo:
                 owner=self.owner,
                 reponame=self.name,
                 branch=branch,
-                path=urllib.parse.quote(directory, safe=""),
+                path=urllib.parse.quote(directory, safe="/"),
             ),
         )
 
--- a/dagshub/upload/wrapper.py
+++ b/dagshub/upload/wrapper.py
@@ -191,7 +191,7 @@ def upload_files(
     remote_path: Optional[str] = None,
     bucket: bool = False,
     **kwargs,
-):
+) -> str:
     """
     Upload file(s) into a repository.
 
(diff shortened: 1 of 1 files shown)


singhvishalkr:
Thanks @coderabbitai -- both catches are valid. Pushed `cee12b6` addressing them:

1. **`new_branch` kwarg**: when `Repo.upload(..., new_branch="branch-x")` is called, the returned URL now uses `branch-x` instead of `self.branch`. Implementation threads `kwargs.get("new_branch") or self.branch` into `get_repo_url(FILES_UI_URL, remote_path, branch=...)`.
2. **`bucket=True`**: there is no canonical file-browser URL for S3 bucket uploads, so the function now returns `None` in that case rather than a misleading repo source URL. Signature changed to `Optional[str]` and the docstring calls this out explicitly.

Ready for re-review.

singhvishalkr:
@coderabbitai thanks for the scan. Pushed `2a191ec` addressing the major URL issue: `get_repo_url` now uses `urllib.parse.quote(directory, safe="/")`, so nested paths such as `data/images/foo.png` render as path separators instead of `%2F`, matching DagsHub's file-browser URL format (`/{owner}/{repo}/src/{branch}/{path}`). The `CONTENT_UPLOAD_URL` caller now benefits from the same behavior.

singhvishalkr:
Coderabbit has re-scanned and the inline suggestions from 2a191ec are all resolved on its side. Ready for a human review whenever someone has cycles.

singhvishalkr:
Latest head should address the earlier Major-tier URL/branch correctness items (see push `2a191ec` and follow-ups). The remaining CodeRabbit note is optional (`None` for bucket uploads vs a canonical `s3://` URI). Happy to add the URI if the team wants that contract; otherwise I would treat it as follow-up and rely on a human pass for merge readiness.

singhvishalkr:
@coderabbitai please re-review the latest head: bucket uploads intentionally return `None`, and repo uploads use `kwargs.get('new_branch') or self.branch` for the FILES_UI URL.

singhvishalkr:
The latest commits are ahead of the CodeRabbit run that left `CHANGES_REQUESTED` (that run compared through `cee12b682361`; head is newer). Nitpick-only feedback was optional S3 URL follow-up.

@coderabbitai please rerun review on head so stale `CHANGES_REQUESTED` clears if nothing blocks.

singhvishalkr:
I checked the current head. Repo.upload() and upload_files() already return the browsable URL for file uploads. I am leaving the optional bucket-upload follow-up untouched for now.

PR #715, opened 2026-08-13
Handle trailing slashes when parsing a repo url in init

### LINKED ISSUE #701: dagshub.init(url=...) with trailing slash sets repo_name="" and repo_owner to the wrong segment
## Problem

`dagshub.init(url="https://dagshub.com/<owner>/<repo>/")` (trailing slash, as copy-pasted from a browser address bar) silently produces `repo_owner="<repo>"` and `repo_name=""`, then calls `RepoAPI("<repo>/")` and `create_repo("", ...)` on the not-found branch.

## Cite

`dagshub/common/init.py` lines 76-79:
```python
if url.endswith(".git"):
    url = url[:-4]
parts = url.split("/")
repo_owner, repo_name = parts[-2], parts[-1]
```
No strip of trailing `/`, no length/emptiness check.

## Steps

```python
url = "https://dagshub.com/owner/repo/"
if url.endswith(".git"): url = url[:-4]
parts = url.split("/")
repo_owner, repo_name = parts[-2], parts[-1]
print(repr(repo_owner), repr(repo_name))
```

## Expected

`'owner' 'repo'`

## Actual

`'repo' ''`

## Environment

- dagshub `0.7.0` (`dagshub/__init__.py`)
- Python `>=3.9` (`setup.py`)
- OS-independent

Fix: strip trailing `/` before split, and raise if `not repo_owner or not repo_name`.

Thanks for maintaining DagsHub/client!

Muhtasim-Munif-Fahim:
Fixes #701.

## The bug

`init(url=...)` takes the last two segments of the url without normalizing it first:

```python
if url.endswith(".git"):
    url = url[:-4]
parts = url.split("/")
repo_owner, repo_name = parts[-2], parts[-1]
```

A url copied from the browser address bar carries a trailing slash, which shifts every segment by one:

| url | `repo_owner` | `repo_name` |
| --- | --- | --- |
| `https://dagshub.com/owner/repo` | `owner` | `repo` |
| `https://dagshub.com/owner/repo/` | `repo` | `""` |
| `https://dagshub.com/owner/repo.git` | `owner` | `repo` |
| `https://dagshub.com/owner/repo.git/` | `repo.git` | `""` |

The `.git` case is worse because the suffix is only stripped when it sits at the very end of the string, so a trailing slash defeats that too.

Nothing catches the empty value afterwards — `RepoAPI("repo/")` is constructed and, on the not-found branch, `create_repo("", ...)` is called.

## The fix

Strip trailing slashes before the `.git` suffix is removed, and fail loudly when the url carries no owner/name pair rather than continuing with empty segments:

```python
url = url.rstrip("/")
if url.endswith(".git"):
    url = url[:-4]
parts = url.rstrip("/").split("/")
repo_owner, repo_name = parts[-2] if len(parts) > 1 else "", parts[-1]
if not repo_owner or not repo_name:
    raise ValueError(...)
```

All four rows above now yield `("owner", "repo")`, as do repeated slashes.

I used `ValueError` for the malformed-url case. `init` already raises `AttributeError` for the mismatched-args case just above, so say the word if you would rather these were consistent.

## Tests

Extends `tests/common/test_init.py` using the fixtures already there:

- `test_init_from_url_tolerates_trailing_slash`, parametrized over the trailing slash, the `.git` + slash combination, and repeated slashes, asserting `RepoAPI` and `create_repo` receive `my-org` / `my-repo`;
- `test_init_from_url_without_owner_and_name_raises`, parametrized over urls with no owner/name pair.

All six fail without the change and pass with it. `tests/common` plus `tests/test_misc.py` is 54 passed.

`tests/common/test_determine_repo.py` fails to collect on my machine for want of `pytest_git`; that is unrelated to this change and reproduces on a clean checkout. I also left the surrounding formatting alone — `black --line-length 120` wants to reformat parts of `test_init.py` that predate this PR, so I kept the diff to the new cases.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/dagshub/common/init.py
+++ b/dagshub/common/init.py
@@ -20,14 +20,39 @@ from dagshub.common.util import lazy_load
 git = lazy_load("git")
 
 
+def _sanitize_repo_url(parsed: urllib.parse.ParseResult, path: str) -> str:
+    """Rebuild a url from its parsed pieces, keeping only scheme, host and path.
+
+    Any userinfo is dropped: `url` is written into MLflow's tracking URI and
+    into .dvc/config, and .dvc/config is a committed file, so a token pasted
+    into the url as "https://user:token@dagshub.com/owner/repo" would end up in
+    the repository. Credentials for both are set separately from the token.
+    """
+
+    host = parsed.hostname or ""
+    if parsed.port:
+        host = f"{host}:{parsed.port}"
+    if not parsed.scheme and not parsed.netloc:
+        # A bare "owner/repo" - nothing to rebuild around.
+        return path
+    return urllib.parse.urlunparse((parsed.scheme, host, path, "", "", ""))
+
+
 def _parse_repo_url(url: str) -> tuple:
-    """Extract (owner, name) from a repository url.
+    """Extract (url, owner, name) from a repository url.
 
     The owner and name are taken from the url *path*, not from the last two
     slash-separated pieces of the whole string. Splitting the whole string lets
     the hostname stand in for the owner: "https://dagshub.com/my-repo" has only
     one path segment, but its last two pieces are "dagshub.com" and "my-repo",
     which looks valid and is not.
+
+    The url is returned alongside them, rebuilt from the same parsed segments,
+    so that the value handed to MLflow and DVC agrees with the owner and name
+    that were derived from it. Parsing alone is not enough: empty segments are
+    skipped when reading owner and name, so without this
+    "https://dagshub.com/my-org//my-repo" would yield the right pair while
+    still sending a doubled slash downstream.
 
     A bare "owner/repo" with no scheme is still accepted, since that is a
     reasonable thing to pass.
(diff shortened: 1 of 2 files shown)


Muhtasim-Munif-Fahim:
Good catch, and it was a real hole — thanks.

`https://dagshub.com/my-repo` has one path segment, but the last two slash-separated pieces of the whole string are `dagshub.com` and `my-repo`, so my check passed it through as a valid owner/name pair rather than raising.

Fixed in e8af285 by parsing the url **path** with `urlparse` and requiring two non-empty segments. With a scheme present the hostname lands in `netloc` and can never be a candidate; without one, `urlparse` puts everything in `path`, so a bare `owner/repo` still works.

Added the reported url and its trailing-slash form to the raising cases. Worth noting the old behaviour made the test suite *hang* rather than fail — the malformed url reached the API instead of raising, which is exactly the failure mode this was meant to prevent.

Muhtasim-Munif-Fahim:
Pushed `6cbd0cf` for both findings.

The url is now rebuilt from the same segments the owner and name are read from, so the two cannot disagree. Previously `https://dagshub.com/my-org//my-repo` reached `RepoAPI` as `my-org/my-repo` while MLflow was pointed at `.../my-org//my-repo.mlflow`.

One thing worth flagging beyond the finding as written: rebuilding also drops any userinfo, and that closes a real leak rather than just tidying the string. `url` is written into `.dvc/config`, which is a committed file — so a token pasted into the url as `https://user:token@dagshub.com/owner/repo` was being committed to the repository. MLflow and DVC are each handed credentials separately from the token, so nothing depended on the userinfo being carried through. The same redaction now applies to the "could not determine the repo owner and name" error, which quotes the url back and is likely to end up in a log.

Five new cases cover this; four of them fail without the change, and the `.git` one passes already and is there as a guard. 104 tests pass in `tests/common`, and flake8 is clean under the arguments the lint workflow uses.

PR #716, opened 2026-08-13
Show complete help text for dagshub repo create

### LINKED ISSUE #310: Truncated command help text for `dagshub repo --help`
Tried running `dagshub repo --help` and this is the full help text that gets printed:

```
Usage: dagshub repo [OPTIONS] COMMAND [ARGS]...

  Operations on repo: currently only 'create'

Options:
  --help  Show this message and exit.

Commands:
  create  create a repo and:
```

Muhtasim-Munif-Fahim:
Fixes #310.

## The listing

click builds a command listing entry from the first line of the docstring. For `create` that line is `create a repo and optionally:`, so the listing ends on a dangling colon and reads exactly as #310 reports — as if the output were cut off:

```
Commands:
  create  create a repo and optionally:
```

An explicit `short_help` on the command makes it a complete sentence:

```
Commands:
  create  Create a repo, optionally uploading data to it and cloning it
          locally
```

## The long help

While reproducing this I found `dagshub repo create --help` is mangled too. The bullet list and the examples are pre-formatted, but nothing told click so, and it rewrapped them:

```
  - upload files to 'data' from a URL dir using `-u` flag. .zip and .tar files
  are extracted,   other formats copied as is.

  example 2:  dagshub --host "https://www.dagshub.com"                 repo
  create mytutorial2 -u "http://0.0.0.0:8080/index.html" --clone --verbose
```

Note the lost indentation on the continuation line and the collapsed run of spaces in example 2. Adding click's `\b` no-rewrap markers keeps the structure:

```
  - upload files to 'data' from a URL dir using the `-u` flag.
    .zip and .tar files are extracted, other formats are copied as is.
  - clone the repo locally using the `--clone` flag.

  Example 1:
    dagshub repo create mytutorial -u "http://example.com/data.csv" --clone

  Example 2:
    dagshub --host "https://www.dagshub.com" repo create mytutorial2
      -u "http://0.0.0.0:8080/index.html" --clone --verbose
```

I treated that as part of the same complaint since it is the same help output, but happy to split it out if you would rather keep the PR to the listing alone.

## Tests

Adds `tests/common/test_cli.py` — as far as I can tell the first CLI test in the suite — using `click.testing.CliRunner` to assert the listing carries a complete summary and that the bullets and examples survive rendering. Both fail without this change.

`tests/common` plus `tests/test_misc.py` is 50 passed. `tests/common/test_determine_repo.py` fails to collect on my machine for want of `pytest_git`, which is unrelated and reproduces on a clean checkout.

Verified against click 8.4.2.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/dagshub/common/cli.py
+++ b/dagshub/common/cli.py
@@ -327,7 +327,7 @@ def repo():
     pass
 
 
-@repo.command()
+@repo.command(short_help="Create a repo, optionally uploading data to it and cloning it locally")
 @click.argument("repo_name")
 @click.option("-u", "--upload-data", help="Upload data from specified url to new repository")
 @click.option("-c", "--clone", is_flag=True, help="Clone repository locally")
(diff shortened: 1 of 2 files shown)


Muhtasim-Munif-Fahim:
Both addressed in 7c55979.

**"URL dir"** — you're right, and I'd carried that wording over from the original rather than checking it. `upload_data` goes straight to `http_request("GET", upload_data)` and the saved name comes from `os.path.basename(urlparse(upload_data).path)`; both examples pass a single file. "dir" invites passing a directory listing URL, which doesn't work. Now reads "upload a file to 'data' from a URL".

**Structural assertions** — also fair. Substring checks would have passed on a rewrap that broke the very layout the `\b` escapes exist to protect, which makes the test close to useless for its stated purpose. It now asserts whole lines including indentation and the full example command, so a formatting regression fails.

